PR #1, opened 2019-02-01
Initial creation of repo

jcontonio:
* Adds basic web and api services with docker-compose
* Adds basic express setup to both services
* Read the README on how to get moving
* Added prettier config and ran it
* Changed all the boilerplate Express Generator stuff to ES6 imports

## New

* Adds React, Webpack, and hot reloading broken out for each EJS view


docker-compose.yml
```diff
@@ -0,0 +1,55 @@
...
+      - ./api:/usr/src/service/api
+    working_dir: /usr/src/service/api
+    command: npm run serve
+    ports:
+      - 3001:3001
+    environment:
+      NODE_ENV: development
+
+
+# version: '3.4'
```
jjidt: do we need this commented out code?

jjidt (dismissed):
this looks like a good start, just the one comment on the commented out code

jcontonio:
This should be good to go now. I've added documentation as to what's going on with Express and Webpack on the web client. You can do local development now with hot reloading, and you can do a production build and start express, getting the hashed webpack bundles.

jcontonio:
Now have webpack configs split up per environment, a simple registration view, and Nodemon / Webpack behaving nicely with each other. Should be a good start.

.gitignore
```diff
@@ -1,3 +1,8 @@
+# App Specific
+web/src/public/css/
+web/src/public/js/
+web/src/public/manifest.json
+
```
dheniges, line 7: Needs `.DS_Store` added, looks like a couple of those files got committed

web/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "webpack-dev-middleware": "^3.5.1",
+    "webpack-hot-middleware": "^2.24.3"
+  },
+  "dependencies": {
+    "cookie-parser": "~1.4.3",
+    "debug": "~2.6.9",
+    "ejs": "~2.5.7",
+    "express": "^4.16.4",
+    "http-errors": "~1.6.2",
+    "morgan": "~1.9.0",
```
dheniges, line 58: there's a few libs in here that may be premature, like the cookie-parser, http-errors, and morgan. Did you copy this config from somewhere or purposefully add these libs?

web/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "node-sass-middleware": "0.11.0",
```
dheniges: node-sass should presumably be a dev dependency. The prod build will want to do the sass builds and make sure those compiled css files are in the right public directory

web/package.json
```diff
@@ -0,0 +1,63 @@
...
+    "react": "^16.7.0",
+    "react-dom": "^16.7.0",
+    "webpack-assets-manifest": "^3.1.1"
```
dheniges: this should also be a dev dependency.

A note on differentiating devDependency vs dependency: they are poorly named as the app still requires some of the devDependencies to build before it's run in prod. Dependencies should be the minimal set of libs required to run the production app _after it's built_. `npm install --production` installs everything in `dependencies`, so that's how we get the prod dependencies and nothing else installed.

web/src/app.js
```diff
@@ -0,0 +1,121 @@
...
+import createError from 'http-errors'
+import { normalizePort } from './libs/express-utilities'
+import { join } from 'path'
+import cookieParser from 'cookie-parser'
+import logger from 'morgan'
+
+// Express Middleware
+import webpackMiddleware  from 'webpack-dev-middleware'
+import webpackHotMiddleware  from 'webpack-hot-middleware'
+import sassMiddleware from 'node-sass-middleware'
```
dheniges: these dependencies won't be available in production, so they can't be imported. I don't remember if the import syntax supports conditional imports or if we'll need to do something else to handle that

web/src/libs/express-utilities.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Normalize a port into a number, string, or false.
+ */
+export const normalizePort = (val) => {
+  var port = parseInt(val, 10)
+  if (isNaN(port)) {
+    // named pipe
```
dheniges: I'm guessing this is copied code from somewhere? The web server is only going to be handling http connections, not named pipes, so this should throw an error to let us know the requested PORT is not a number to make config validation/debug easier

web/webpack.config.js
```diff
@@ -0,0 +1,90 @@
+const webpack = require('webpack')
+const path = require('path')
+const MiniCssExtractPlugin = require('mini-css-extract-plugin')
+const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin')
+const WebpackAssetsManifest = require('webpack-assets-manifest')
+
+const devMode = process.env.NODE_ENV !== 'production'
```
dheniges: We're going to want to break out this webpack config into either 2 files, or import config for the requested environment. The problem with the ternary expressions sprinkled throughout the file is what happens when you want to add a 3rd environment? Our local dev config may be different from a kubernetes dev environment which may be different than a CI environment etc. etc.

This can be done later.

dheniges (changes requested):
Nice job wrangling this all together. There's a lot of things to think about starting a new app, thus the deluge of comments. Happy to help out on any of the requested updates.

web/webpack.config.js
dheniges: Apparently I didn't review the very latest, as you already broke out the webpack config into two files 👏   I just browsed through the files again and looks like everything else is set, not sure how I didn't have the latest on the webpack config files

dheniges (approved):

PR #2, opened 2019-02-04
Production builds for web server

dheniges:
- Preprocess web server (ie node.js) files with babel
- Move server files into their own directory so it's easy to know what needs processing
- Move webpack bundles into dist/public so we know exactly what we're exposing publicly
- Update paths everywhere to achieve the above

The only piece I don't see working is the css but that's a known issue from removing the node-sass middleware


jcontonio:
Re: Sass, add a `&& node-sass` CLI command to the build script prior to running the server

jcontonio (changes requested):
1. IMO moving server to the root is messy. What do you think about moving server into the src directory? I'd prefer a src and dist at the top level, and not much else. I often only open the "src" folder when doing work, so that only the things I care about are indexed, searchable, etc.
2. The public directory shouldn't be there at all, or checked into git. It seems I ignored everything in it, but kept it around. It should only exist when running build and getting generated in the dist directory.

So a structure like this in the repo - and `dist` being what everything builds into

```
/tools
  ..config files for things..
/src
  /components
  /libs
  /server
    /routes
    /views
    /scripts
    /css
    /images
    app.js

jcontonio:
I'm also beginning to think, maybe webpack should bundle all of the CSS and images for production build, including the "server" assets so that it can hash those as well... I can do that in another PR later this week.

dheniges:
The public directory is just a .gitkeep, all the actual public files are ignored. That way the images always have the directories and we don't have to worry about whether they exist or not which can cause unexpected problems

jcontonio:
We can delete this now?

jcontonio (approved):
Let's merge

PR #3, opened 2019-02-05
Production builds + Kubernetes

dheniges:
- Modifies prod builds PR to kubernetify things
- API will need updates to do the same things, but currently runs fine (just not in a hardened prod mode)
- Removed some unused files
- Added a `run.sh` entrypoint for the web image to standardize basic commands across services

Prod images can now be deployed with the kubernetes config. The kubernetes folder can be moved into its own ops repo later if that makes things easier to manage, including config changes that don't require rebuilding images


jcontonio (approved):
👍

PR #4, opened 2019-02-06
stripe api

jjidt:
TO TEST: use credit card number 4242424242424242

- sets up basic api structure
- sets up basic stripe api usage
- adds dotenv required from cli so we can pass in stripe secret and api url
- adds temporary web component with basic sign up form (note some things in this are hardcoded such as api url, which we can tackle when the web structure is locked and add dotenv there as well)
- babel config for api
- sets up controller/coordinator/integrator file structure
----controllers will handle validation and http responses/higher level error handling
----coordinators will handle cases where multiple areas are accessed in a single process (in this case stripe integrator and user model)
----integrators will wrap communication with a single outside service

I would also like to throw out using a repository layer on top of whatever ORM we choose

these things together will make testing easier and swapping out libraries less painful


dheniges (commented):
Weirdly no feedback on the code :)

Since things are moving around in web so much I want to review Jay's PR and merge that first, and then move files around as needed in this PR on the web-side.

api/.babelrc
```diff
@@ -0,0 +1,12 @@
+{
+  "presets": [
+    [
+      "env",
+      {
+        "targets": {
+          "node": "10.15.1"
```
dheniges: I lied about code feedback. I learned from Jay's PR this can be set to `current` and will use the version of node running. This way we don't have to update the node version in multiple spots

dheniges (changes requested):

dheniges:
@jjidt can you update this PR to move the web server stuff into the new directory structure?

dheniges (approved):

PR #5, opened 2019-02-06
Updated webpack bundling with express middleware

jcontonio:
* Moves all client stuff into web/src/client
* Updates Express app.js to create bundle objects for prod and dev
* Hot reloading works for style-loader based sass


web/.babelrc
```diff
@@ -1,4 +1,4 @@
 {
   "plugins": ["@babel/plugin-syntax-dynamic-import"],
-  "presets": ["@babel/preset-env", "@babel/preset-react"]
+  "presets": [["@babel/preset-env", { "targets": { "node": "current"}}], "@babel/preset-react"]
```
dheniges, line 3: that "current" target is nice 👍

web/src/client/components/SampleComponent/SampleChild.js
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 
 const SampleChild = () => (
   <div className="c-sample-child">
-    <h1>Sample aaa</h1>
+    <h1>Sample JJ</h1>
```
dheniges, line 5: Quality update

web/src/client/components/SampleComponent/index.js
```diff
@@ -4,6 +4,10 @@ import SampleChild from './SampleChild'
 import './SampleComponent.scss'
 
 class AppContainer extends React.Component {
+  constructor(props) {
+    super(props)
+    console.log('s')
```
dheniges, line 9: Is this for anything? I don't really care since this is a dummy component but wondering if it's on purpose

web/config/webpack.dev.js
```diff
@@ -1,29 +1,47 @@
...
+    }
+  }
+]
 
 module.exports = {
   mode: 'development',
   entry: {
-    index: entryPoints['index']
+    index: entryPoints['index'],
+    registration: entryPoints['registration']
```
dheniges: can you just pass `entry: entryPoints` here? Looks like it's recreating the entryPoints object that already exists

web/config/entryPoints.js
```diff
@@ -0,0 +1,13 @@
+
+const middlewareScript = 'webpack-hot-middleware/client?path=/__webpack_hmr&timeout=20000&name='
```
dheniges, line 2: This path is getting included in prod. If this is the only entrypoint that's different between dev/prod, you could just append this script to the object below:
```
const exports; // current exported object
if (process.NODE_ENV === 'development') {
  exports.index.push(middlewareScript);
}
module.exports = exports
```

dheniges (changes requested):
A few tweaks to entryPoints with webpack and I think this is good to go

web/src/client/components/SampleComponent/index.js
jcontonio: No, that was me testing hot reloading

web/config/entryPoints.js
jcontonio: Good catch

dheniges (approved):

PR #6, opened 2019-02-06
Docker prod build + dev dependency updates

dheniges:
- Docker npm install order change for more efficient use of caching (ie faster builds)
- Explicit process for clearing dev node_modules volume when new dependencies added


jcontonio (approved):
👍

PR #8, opened 2019-02-07
linting

jjidt:


api/package.json
```diff
@@ -6,23 +6,41 @@
...
-    "nodemon": "^1.18.9"
+    "eslint-config-prettier": "^4.0.0",
+    "eslint-plugin-prettier": "^3.0.1",
+    "husky": "^1.3.1",
+    "lint-staged": "^8.1.3",
+    "nodemon": "^1.18.9",
+    "prettier": "^1.16.4"
   },
   "dependencies": {
+    "eslint": "^5.13.0",
```
dheniges: should be a dev dependency?

dheniges:
From JJ: "looks like the editor is picking up some linting issues not found during commit, if you wanna take a look at those globs in the package.json lint-staged config for web and api, maybe I'm missing something"

dheniges (dismissed):
The husky precommit hook isn't triggering for me, so it's not running `eslint --fix`. Is there additional setup/config needed for that?

Also, starting/restarting the API is real slow with nodemon. I wouldn't expect the stripe dependencies to be massive. Not seeing anything obvious in babel-node, which could be a culprit

web/package.json
```diff
@@ -22,6 +22,23 @@
     "maintained node versions",
     "not dead"
   ],
+  "lint-staged": {
+    "linters": {
+      "*.{js,jsx,css,md}": [
```
dheniges: this should use the same pattern dir match as the api: `"src/**/*.{js,jsx...}"` but with css and whatnot

dheniges (dismissed):
I updated the package config for lint-staged to use relative paths to the package.json file. lint-staged is a pretty nice lib in that it has a debug tool built-in to the binary: https://github.com/okonet/lint-staged#command-line-flags

Two remaining issues:

1. On web there are conflicting rules between prettier and eslint about parens around a single argument in a function. Running eslint --fix will switch back and forth between adding or removing parens.

2. There's an empty .eslintrc file in the root of the repo that should be removed

jjidt:
@dheniges check it again, I tested with a bunch of files and it seems to be catching things correctly now

runs prettier via pretty-quick on staged files before linting, so bypasses the arrow paren nonsense

api/package.json
```diff
@@ -6,21 +6,42 @@
...
   },
   "repository": {
     "type": "git",
     "url": "git+ssh://git@github.com/BonkeyBong/project-danger.git"
   },
   "author": "BonkeyBong, LLC",
   "license": "ISC",
+  "lint-staged": {
+    "linters": {
+      "*.{js}": [
```
dheniges, line 20: this rule is not picking up changed files for me in the server.

screen shot 2019-02-07 at 1 39 31 pm

That relative path I added before fixed this. For reasons I don't understand at all, that config works in web, but I like the explicit relative path config so the web linter can never pick up changed API files.

Weird web config below that's working:
screen shot 2019-02-07 at 1 41 48 pm

dheniges (changes requested):
The eslint/prettier weirdness seems resolved!

But, API changes are not getting detected based on the rule. It works in web with the same rule which makes no sense

The config I added earlier fixed this for me:
```
    "linters": {
      "src/**/*.{js,json}": [
        "eslint --fix",
        "git add"
      ]
    },
    "relative": true
```

Lint debug run with the above settings:
screen shot 2019-02-07 at 1 46 47 pm

jjidt:
using 'relative' flag on api settings is resulting in this for me when I change a file and try to commit

```fatal: pathspec 'src/coordinators/billing.js' did not match any files
husky > pre-commit hook failed (add --no-verify to bypass)

dheniges (approved):

dheniges:
For posterity:

For reasons unknown the lint-stage config in api/package.json is not triggering the linting on the post-commit hook for me. Exact same config works fine in web 🤷‍♀️

PR #9, opened 2019-02-07
Docker/build updates

dheniges:
Dependent on #8 

- Node alpine versions (`danger/web` 212mb instead of previous 1.03gb!)
- Full prod build process for API now
- API env var validation
- Optimizations to web Docker build
- Dev hookup for docker-compose with API


jcontonio (approved):
👍 Works great - Hot reloading and all

PR #11, opened 2019-02-08
UI Kit implemented - Updated Stripe form

jcontonio:
* Added new routes for sign up vs billing
* Created Message component for flash messages
* Added shared directory for constants
* Broke up views into partials


web/src/client/js/billing.js
```diff
@@ -0,0 +1,92 @@
...
+// Create a token when the form is submitted.
+var form = document.getElementById('payment-form')
+form.addEventListener('submit', function(e) {
+  e.preventDefault()
+  createToken()
+})
+
+if (process.env.NODE_ENV !== 'production' && module.hot) {
+  module.hot.accept()
+}
```
dheniges: Instead of having to add this module.hot.accept() in each script, can the header or footer partial include a script that just does this?

web/src/client/js/index.js
```diff
@@ -1,8 +1,8 @@
-import React from 'react'
-import { render } from 'react-dom'
-import SampleComponent from '../components/SampleComponent'
+// import React from 'react'
+// import { render } from 'react-dom'
+// import SampleComponent from '../components/SampleComponent'
```
dheniges: Delete?

dheniges (changes requested):
Looks good, two minor comments

web/src/client/js/index.js
jcontonio: Yeah

web/src/client/js/billing.js
jcontonio: Fixed with a global JS file for all entryPoints, removed the index one

dheniges (approved):

PR #12, opened 2019-02-09
billing unit tests and api test setup

jjidt:
sets up jest/ some basic tests / and mocking pattern


api/src/coordinators/billing.test-u.js
```diff
@@ -0,0 +1,41 @@
...
+      expect(stripeIntegrator.createCustomer.mock.calls[0][0]).toEqual({
+        email,
+        token
+      })
+    })
+
+    it('should call stripeIntegrator.createSubscription with the customer id and plan id', () => {
+      expect(stripeIntegrator.createSubscription.mock.calls.length).toBe(1)
+      expect(stripeIntegrator.createSubscription.mock.calls[0][0]).toEqual({
+        customerId: stripeIntegrator.createCustomer.mock.results[0].value.id,
```
dheniges, line 32: What's the quick overview of how jest is doing these mocks? The `.calls[0][0]` is throwing me off, and this `mock.results` is pulling the result from the mock call on 31?

dheniges (changes requested):
Setup looks good, just a question on the jest mock-age.

Only requested change is adding `**/__mocks__` to the API `.dockerignore` file, as well as the regex for test files so they don't make it onto the prod instance (not that it should matter, but might as well keep it clean)

api/src/coordinators/billing.test-u.js
jjidt: .calls[0][0] references the first call, first arg

On line 32 I'm verifying that the results of the createCustomer mock are getting passed into the createSubscription mock, but I'm pulling that directly from the mock itself rather than hardcoding the value. Pretty much, if I change the return value in the createCustomer mock down the road, I don't want to have to go back and update hardcoded values in test files.

jjidt:
updated, I included the generic default test regex and also a custom one that will include test-u test-i test-a files

dheniges (approved):

PR #13, opened 2019-02-11
set up client testing

jjidt:


dheniges (changes requested):
Test setup looks good 👍 

Same comments on ignoring files:
- Add the coverage directory to .gitignore and .dockerignore
- Add the test file regex to .dockerignore (and mocks, or we can wait until there are actually mocks)

jjidt:
@dheniges check those dockerignore changes to see if they look valid

the coverage directory was already gitignored in the main repo file

web/.dockerignore
```diff
@@ -1,4 +1,6 @@
 node_modules/
 .eslintrc
 .prettierrc
-coverage
\ No newline at end of file
+coverage
+(/__tests__/.*|(\\.|/)(test|spec))\\.[jt]sx?$
+*/*/__mocks__
```
dheniges: `**/__mocks__` will ignore mock folders at any subdirectory depth

dheniges (approved):

PR #14, opened 2019-02-12
Editor Prototyping

jcontonio:
* Adds content editor at http://localhost:3000/dashboard
  * Try writing in markdown for italics, bold, and headings and things like that
* Adds password reset, more/smarter partials, and adding meta tags / images to quiet the 404's

I need help with the RegEx's. You'll see that the bold tags are getting hi-jacked by the italic pattern. It's in EditorComponent.js, line 42


jcontonio:
image

web/src/shared/utilities.js
```diff
@@ -0,0 +1,24 @@
+export const debounce = function(delay, fn) {
```
dheniges, line 1: Don't think this is used anywhere?

dheniges (dismissed):
Looks good except maybe one unused function?

If we switch editors that's cool, but might as well have something in there at the moment

web/src/shared/utilities.js
jcontonio: It's not used, but we will definitely use it - should leave it in

jjidt (approved):
👍

PR #15, opened 2019-02-12
API auth

dheniges:
- Local login setup with dummy username/pass for now
- JWT auth setup for authenticated endpoints
- Controller loader to apply middleware. Makes adding/removing/updating middleware changeable in a single spot

### DIFF
--- a/api/src/controllers/login.js
+++ b/api/src/controllers/login.js
@@ -8,7 +8,7 @@ const loginController = function(router) {
   router.post('/', auth.loginMiddleware, login)
 }
 
-const login = async function(req, res) {
+const login = function(req, res) {
   const token = tokenIntegrator.generateToken(req.user.email)
   res.json({
     token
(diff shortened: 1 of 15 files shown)


api/src/controllers/login.js
```diff
@@ -0,0 +1,18 @@
...
+import auth from '../libs/auth'
+import tokenIntegrator from '../integrators/token'
+
+// Caution: login controller routes have NO AUTH by default!
+
+const loginController = function(router) {
+  router.post('/', auth.loginMiddleware, login)
+}
+
+const login = async function(req, res) {
```
jjidt: doesn't look like this function needs to be async, or is that planning ahead for a future implementation?

jjidt (dismissed):
looks/works great!
Just the one comment, fix if necessary, but if not merge away 👍

api/src/controllers/login.js
dheniges: hah, no, not planning ahead, just getting a little async happy

dheniges:
@jjidt still need your official review approval before it's mergeable. Fixed some merge conflicts and removed the async

jjidt (approved):
👍 nice

PR #16, opened 2019-02-12
api validation

jjidt:
- sets up basic validation middleware using joi
- note that we don't want to compile the joi schemas at runtime, so it's pre compiled at the module level
- also note that we're not logging any data beyond whatever 400 response logs will get generated by express
- also the current settings will cast to the type and mutate the response body, query, etc. if "1" is passed to a field wanting a Joi.number() it will be cast to a number
- we may want to lock this down to just body, request, and headers on the request object, but I think it's ok to leave open ended for now


dheniges (approved):
This looks quite nice, I haven't used joi before

PR #18, opened 2019-02-19
initialize db & orm

jjidt:
- sets up postgres via docker compose
- sets up sequelize connector module and initializes before starting server
- integration tests for db connection

`make test-api` will set up docker db and run api tests


Makefile
```diff
@@ -13,3 +13,11 @@ start:
 
 stop:
 	docker-compose stop
+
+test-api:
+	docker-compose down && \
```
dheniges: Is this just to make sure that docker-compose isn't already running? The `down` wipes out the volumes which just makes it slower to restart as it'll npm install everything again. Not sure running tests requires that?

api/wait-for.sh
```diff
@@ -0,0 +1,80 @@
...
+}
+
+wait_for() {
+  for i in `seq $TIMEOUT` ; do
+    nc -z "$HOST" "$PORT" > /dev/null 2>&1
+    
+    result=$?
+    if [ $result -eq 0 ] ; then
+      if [ $# -gt 0 ] ; then
+        echo "SUCCESSFULLY CONNECTED"
```
dheniges: maybe preface this with "WAIT_FOR.sh" or something to make debugging from container logs easier

dheniges (changes requested):
Feedback:
- We'll want the ability to execute tests within docker containers, as this will eliminate the need for tests dependencies on whatever CI service we use. You can see the run.sh file in api/ will take a `test` command to run `npm test`. You should just be able to modify the existing docker-compose test file
- `make start` is kind of broken with this setup now because there are no default credentials for dev to use to connect to the db. You can either add them to the docker-compose file for dev, or in my PR I've updated docker-compose to use the .env file, in which case you could add the default settings in example.env

jjidt:
updated
removed `down` but it seems like we need the build step in the test script correct?
on a docker restart, it would fail until an initial build ran

dheniges:
yep, build step is still required. My previous comment was referencing the fact the docker volumes will persist if a build was already run and the npm install step will be much quicker

dheniges (changes requested):
Updates to dev .env example looks good, I can start the API and it connects to the db 👍 

I'm not able to run `make test-api` successfully, the tests are invoked in the API, but I don't think credentials are setup for the API to connect to the db with the docker-compose.test.yml file. For now I think adding them to the compose file is simplest.

dheniges (approved):

PR #19, opened 2019-02-19
Web login auth

dheniges:
Bunch of changes to get web logins using the API:
- .env support and handling with docker-compose and the web
- Note updated readme to copy example.env in web/src/server/
- Basic API request handler lib for the web server. This won't be our SDK as it needs to not store auth state as requests are made on behalf of a multitude of users
- Current placeholder render of the user object with the token after successful login. Figured this was a good stopping point

To avoid session management, I think we can put the token in a cookie, I'll work on that next

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@ To get started, you'll need Docker installed, then run the following commands
 ```
 // Local development
 cp api/example.env api/.env
+cp web/src/server/example.env web/src/server/.env
 make builddev
 make start
 make stop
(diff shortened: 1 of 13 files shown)


jcontonio (changes requested):
Looks good, could log in 👍 Should we add a /sign-out while we're at it? Only change is making the `cleardev` change discussed in chat

dheniges:
Per slack, I realize I don't want to combine the builddev and cleardev steps. I like the idea of a "rebuild from scratch" option later.

As for sign-out, there's no login state stored anywhere so there's nothing to do on that route yet

jcontonio (approved):

PR #20, opened 2019-02-19
Container / Component structure + Redux and Routing

jcontonio:
* Adds React Router and Redux
* Separates Containers (app sections) and Components (reusable components)
* Adds webpack config for resolving directories so we don't have to worry about paths to a container or component

Example, these will resolve anywhere 

```
import BlahContainer from 'Containers/Blah'
import BlahComponent from 'Components/Blah'
```

I need some help thinking through Express's routes, webpack's bundle, and React Router... Take a look at Express, in app.js line ~22


jcontonio:
What about if we remove the "dashboard" route file all together, and handle it in index.js, and render the dashboard view if we're logged in. 

```
router.get('/', (req, res, next) => {
  if (authd) {
    res.render('index', renderIndexBundles(req, 'Home'))
  } else {
    res.render('app', renderAppBundles(req, 'Dashboard')
  }
})
```

Does that solve everything?

dheniges:
Looks like you are wanting a better way to specify a list of routes on the web server that will return the client app bundle?

From my end, I'd like a clear/obvious way to say what routes are handled by the client app and what routes are handled by the express app.

The auth question is comes down to:
1) does the route require authorization? (I would potentially argue any route loading the client app bundle requires auth)
2) Is the user authenticated? If not, express middleware can redirect them wherever we want. eg `/home`

The app.js file is already a bit monstrous, so keeping route definitions somewhere else would be my preference. In the API you can see my approach to loading controllers and inserting auth middleware, I think I'd like to take a similar approach on the web server so adding middleware is easy and can be done in one spot for groups of controllers. This also leads toward the solution mentioned above where we can define the routes that should load the client app bundle:

```
const CLIENT_APP_ROUTES = ['/dashboard', '/projects', '/project']
const middlewares = [auth.jwtMiddleware] // whatever future middlewares me want on these routes

// insert _after_ all the other non-route-specific middleware!
CLIENT_APP_ROUTES.forEach((path) => {
  app.route(path).all(...middlewares, (req, res) => {
    return render('bundle')
  })
})

```

dheniges:
I'll finish reviewing tomorrow, but code looks good so far

web/src/client/js/containers/Projects/Projects.js
```diff
@@ -0,0 +1,54 @@
...
+  projects: PropTypes.object
+}
+
+const mapStateToProps = (state) => {
+  return {
+    projects: state.projects
+  }
+}
+
+// const mapDispatchToProps = (dispatch) => {
```
jjidt, line 46: do we need this commented code here?

web/src/client/js/reducers/project.js
```diff
@@ -0,0 +1,29 @@
+import { ActionTypes } from '../actions'
+
+// import { ActionTypes } from '../actions'
```
jjidt: looks like a duplicate

jjidt (changes requested):
couple cleanup things, other than that looks like a good start for react/redux structure:

overall I agree with what dirk was saying about splitting routing to a `post-login` grouping where the react app bundle will be delivered and routed client side

and a `pre-login` grouping which can be however many different express routes we need (marketing, sign-up, docs, etc.) and templated however we want

web/src/client/js/containers/Projects/Projects.js
jcontonio: We don't need it, but it was there to remind me that that function exists for when we get to that point. I didn't realize you could map them like that with Redux...in Gates, we're manually calling all those actions. We could remove it.

jjidt: cool, I'm into leaving it as a placeholder, just wanted to make sure it was something we'll want

web/src/client/js/reducers/project.js
jcontonio: 👍

jjidt (approved):
nice 👍

PR #22, opened 2019-02-22
Billing component & better bundle loading

jcontonio:
Bunch of work here to get `<BillingContainer />` to be usable outside of the app bundle or in it. 
You can check:

http://localhost:3000/billing 
http://localhost:3000/d/dashboard

**Also:**

* Rewrote the `BundleBuilderPlugin` that Express is using to manage the dev bundles. 
* Did a diff of the webpack configs and combined as much as the similar stuff as I could.
* Added autocomplete stuff to form fields
* Added initial service worker stuff for caching and offline functionality
* Cleaned up the routes to use one function for including bundles

I tried to run the production build, but don't really know how. I tried doing a production build of the client, and running `npm start`, but I am getting the following error from the client API library there. Doesn't seem related to this PR:

```
➜  web git:(jc-billing-component) npm start

> project-danger-web@1.0.0 start /Users/jcontonio/Sites/project-danger/web
> NODE_ENV=production node ./dist/server/app.js

/Users/jcontonio/Sites/project-danger/web/dist/server/libs/api/index.js:27
    if (typeof url !== 'string' && url.length < 1) {
                                       ^

TypeError: Cannot read property 'length' of undefined
    at new DangerApi (/Users/jcontonio/Sites/project-danger/web/dist/server/libs/api/index.js:27:40)
```


dheniges (approved):
👍

PR #23, opened 2019-02-27
Web server auth

dheniges:
Web server full auth flows

**Note** you'll need to update .env in web and maybe api

- Moves controllers into controllers folder
- Adds cookie with jwt token when user signs in
- Protects controllers with cookie jwt auth unless explicitly added to unauthorized routes
- Allows url aliases (aka vanity urls) for any route. Add route alias to `/libs/aliases` (currently only sign-in is aliased)

Next step is pulling the jwt out of the cookie in the client app so it can be used to make API requests


jcontonio (approved):
Can log in and log out, looks good 👍

PR #25, opened 2019-03-02
User model

dheniges:
Builds off JJ's work to create the user model and schema.

- Adds database migration and seeds
- Starting the dev environment will run database migrations and seed data

Need to lock in a direction on the BusinessTimeLayer™, currently it's dumb and only there to function with login.

Also need to see if there's a way we can do something with the compiled bcrypt binary for alpine linux so we don't have to add all the build tools to the image, not to mention waiting for it to build


api/src/controllers/billing.js
```diff
@@ -19,7 +20,10 @@ const billingController = function(router) {
 }
 
 const getBilling = async function(req, res) {
-  res.json({ billing: 'loaded' })
+  const db = getDb()
```
jjidt: looks like some test code I left in there, wanna delete this?

jjidt (changes requested):
Looks good 👍 super clean and easy to follow, one cleanup thing that was left in there then it'll be good to merge

couple thoughts:
if we change businesstime layer to only pass out 'plain old javascript objects', we'll need to add user lookup to the 2 password functions

we might want to instantiate the db in module scope and have it directly accessible from the import, so that we don't always have to call getDb() (on the fence about this one)

dheniges:
@jjidt updated. Not doing anything real with the BusinessTimeLayer™ yet. As you pointed out, not sure how we'll want to handle those scenarios

jjidt (approved):

PR #26, opened 2019-03-04
Client app token

dheniges:
- Client app parses out token
- Added scaffolding for initial API lib in client app
- Builds off #25


dheniges:
Superseded by #28

PR #28, opened 2019-03-06
Projects in API and client

dheniges:
- Adds create/view/list endpoints for projects to API
- Lists projects in the client
- Uses webpack.DefinePlugin to do static replacement of variables with environment variables
- Fixes CORS handling in API

There are currently no permissions, few validations etc. applied on these endpoints, still got a bunch of work to do there


jcontonio (approved):
👍

PR #29, opened 2019-03-07
ability to execute sequelize commands via makefile

dheniges:
- Now we can sequelize all the things! The `make sequelize` command shows all the options, everything is pretty self-explanatory
- Doc with specific scenarios added for seed data and migrations


jcontonio (approved):
Works great 👍

PR #30, opened 2019-03-08
UI Header and Navigator component – Fetch Project

jcontonio:
* Moves old containers into new pure components that render markup + connected containers
* Makes each component self-sufficient, fetching their own resources, getting the params props through withRouter - goal is to just plop in components wherever, and they work
* Adds new Header with different modes (/d/dashboard, /d/projects, /d/project/1)
* Starts new CSS system for the app (separate from the front-page / marketing stuff)
* Removes Bulma from any app CSS
* Adds base setup for branding coming from the organization in the header
* Adds fetching of Project, and Project redux state


Makefile
```diff
@@ -4,9 +4,8 @@ build:
 builddev:
 	./build.sh && docker-compose build
 
-cleardev:
-	# Erases the volumes, including node_modules. Run this if node_modules changes
-	docker-compose down
```
dheniges: We still need the cleardev to erase volumes, running build, even with no cache, will still mount the existing volume

Makefile
```diff
@@ -4,9 +4,8 @@ build:
...
+rebuilddev:
+	./build.sh docker-compose build --no-cache
```
dheniges: This command doesn't work, it's running the build script and passing docker-compose as args that just get ignored

web/src/client/js/components/Navigator/NavigatorContainer.js
```diff
@@ -0,0 +1,108 @@
...
+    this.nodeRef = React.createRef()
+    this.menuRef = React.createRef()
+    this.state = {
+      expanded: false
+    }
+  }
+
+  componentDidMount() {
+    if (Object.keys(this.props.projects).length === 0) {
+      // If we don't have any projects, go get them
```
dheniges, line 18: This works for now, but we'll need a solution that handles possible empty returns in the future. I think that work falls under the fetch once from API task in Asana

web/src/client/js/reducers/projects.js
```diff
@@ -0,0 +1,34 @@
...
+        projectsById[project.id] = project
+        return projectsById
+      }, {})
+      break
+    default:
+      return state
+  }
+}
+
+export const project = (state = initialState, action) => {
```
dheniges: If these are two different stores they should be in separate files to keep things less confusing. Right now they're sharing the same initialState for instance, which we'll never want.

dheniges (changes requested):
Comments on the container pattern:

The NavigatorContainer is mixing data fetching + component logic into one component. The container pattern exists to gather data from wherever and feed it to the component. This gives us two things:

1) Separation of concerns. Data is always handled in one spot, and display is always handled in one spot. When we figure out a nice flow for data fetching, it'll plugin at the container layer and not impact anything else. It also eases maintainability as our list of components grows.

2) Reusability. In the case of the Navigation this may not apply, but consider a list component that can display projects, documents, or a combination. By feeding data to the component from a container, we can reuse that component but with different containers to show different lists in different spots (perhaps a search results list here, and a recent projects list there)

So for the NavigatorContainer, all it would do is fetch projects, and have the mapStateToProps and mapDispatchToProps. It would render a single child component, <Navigator />, passing the necessary props. Navigator would hold the internal state and logic for opening/closing the menu via the refs, as currently exists.

The ProjectContainer is the perfect example of a separated data fetching layer. Right now it just has a dummy render of the project name, but going forward that render will presumably be a <Project /> component that takes a project object prop

Makefile
jcontonio: Oh woops, forgot the &&

dheniges (changes requested):
functionality is good, last change is in the makefile adding `docker-compose down && ./build.sh ...` to clear the volumes when doing a rebuilddev

dheniges (approved):

PR #31, opened 2019-03-09
Hooks! Converts NavigatorComponent to use React Hooks

jcontonio:
* Creates a few custom hooks for common UI listener / actions
* Converts NavigatorComponent to be a React Function, using hooks and a reducer to manage state
* Adds eslint plugins for react-hooks and jsx-a11y for accessibility
* Makes the NavigatorComponent accessible

### DIFF
--- a/web/src/client/js/components/Navigator/NavigatorComponent.js
+++ b/web/src/client/js/components/Navigator/NavigatorComponent.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef, useReducer } from 'react'
+import React, { useState, useCallback, useEffect, useRef } from 'react'
 import PropTypes from 'prop-types'
 
 import useClickOutside from 'Hooks/useClickOutside'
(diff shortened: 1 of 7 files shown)


web/src/client/js/components/Navigator/NavigatorComponent.js
```diff
@@ -1,77 +1,63 @@
...
 
-  renderTitle() {
-    const { match, project } = this.props
-    const section = match.path.split('/')[1]
-    return match.params.projectId && section === 'project' ? project.name : 'Projects'
-  }
+  // If project Id changes, trigger contract
+  useEffect(() => {
+    dispatch({ type: 'contract' })
+  }, [project])
```
dheniges: the docs don't say how React is going to do object comparison. Presumably it's a shallow comparison?

web/src/client/js/hooks/useBlur.js
```diff
@@ -0,0 +1,26 @@
+import { useEffect } from 'react'
+
+function useBlur(ref, dispatch, action) {
+  let timeOutId
```
dheniges: I think this variable should be created via `useState`? I'm not 100% certain, but since you're setting this variable in callbacks in `useEffect`,  I think you want to use the setter provided by useState and not rely on the closure. Because this useBlur function is going to get called anytime the implementing component changes at all during its lifecycle.

web/src/client/js/hooks/useBlur.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  useEffect(() => {
+    ref.current.addEventListener('focusout', blurHandler, false)
+    ref.current.addEventListener('focusin', focusHandler, false)
+    return () => {
+      ref.current.removeEventListener('focusout', blurHandler, false)
+      ref.current.removeEventListener('focusin', focusHandler, false)
+    }
+  })
```
dheniges: I think you want a second empty array arg here to `useEffect` so these listeners are only added when the component first mounts. https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect See the note above the yellow box about the empty array

web/src/client/js/hooks/useClickOutside.js
```diff
@@ -0,0 +1,27 @@
+import { useState, useEffect } from 'react'
+
+function useClickOutside(ref, dispatch, action) {
```
dheniges: I noticed this in your other custom hook, `dispatch` and `action` are passed as two separate args, but it's always invoked as `dispatch(action)`. Seems like it'd be more generic to have an event handler callback arg for the hook to invoke instead, and the component can pass a function that does the dispatch:
```
`useClickOutside(nodeRef, () => { dispatch(action) }`
```
This hook also seems to be using two patterns: it's returning a true/false click value as well as invoking a callback. It seems like returning the click is a simpler to follow approach.

web/src/client/js/components/Navigator/NavigatorComponent.js
```diff
@@ -1,77 +1,63 @@
...
-  }
-
-  expandMenu() {
-    this.setState({ expanded: !this.state.expanded }, () => {
-      this.menuRef.current.hidden = !this.state.expanded
-    })
-  }
+  // The title of the component should be Projects or Project Name
+  const [title, setTitle] = useState('')
+  useEffect(() => {
```
dheniges: I don't think this is a `useEffect` scenario, as the title is a simple variable based on the project name. Effects are basically anything you'd add in a lifecycle hook on a React class, like in `componentDidMount` and similar. This is a straightforward render variable here and could be declared as such

dheniges (commented):
There are a bunch of comments on hook implementation. I'm still learning the API and best practices so let's walk through this tomorrow with @jjidt so we can all align ourselves on hooks.

dheniges (approved):
Yaaaay hooks

PR #32, opened 2019-03-11
Resource view abstraction for React components

dheniges:
- Implements a new React hook, `useDataService`
- Fetches and subscribes to resource views requested by a component
- Supersedes react-redux's `connect()` HOC
- Resource views defined in dataMapper file
- Implemented in ProjectsContainer


dheniges:
Some thoughts so far:
- by having the resource "views" defined as a function in dataMapper.js, args can be passed for filters, pagination etc. and usable by the fetchAction, getState... etc. functions
- it may be desirable to have `useDataService` return an object or array so the loading status can be passed to the component. Currently the component has to know what the initial state in the store is vs. the loaded state to "guess" loading status. Currently `useDataService` just returns the data

jcontonio (approved):
👍

PR #33, opened 2019-03-12
businesstime!

jjidt:
- moves user and project fetching/creation to businesstime layer
- moves user password functions to coordinator layer since this is touching the user model and also accessing the password generator/validator
- adds ono library for Error creation, which allows us to easily set error.code property anywhere in code, and then send that status to the api consumer.  Also allows us to re-throw errors with updated messages/data without losing stack trace
- All data coming out of BusinessTime :tm: are now POJOs

### DIFF
--- a/api/src/controllers/projects.js
+++ b/api/src/controllers/projects.js
@@ -1,6 +1,7 @@
 import Joi from 'joi'
 import validate from '../libs/payloadValidation'
-import { getDb } from '../db/db-connector'
+import BusinessProject from '../businesstime/project'
+import ono from 'ono'
 
 const projectsPayloadSchema = Joi.compile({
   body: Joi.object().keys({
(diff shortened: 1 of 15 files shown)


api/src/controllers/projects.js
```diff
@@ -1,61 +1,56 @@
...
   router.get('/:id', getProject)
 }
 
 const getProjects = async function(req, res) {
-  const db = getDb()
-  let projects
   try {
-    projects = await db.model('Project').findAll({}, { raw: true })
+    const projects = await BusinessProject.findAll()
+    throw new Error('blarp')
```
dheniges: v. informative

api/src/coordinators/user.js
```diff
@@ -0,0 +1,25 @@
+import passwords from '../libs/passwords'
+import userBusiness from '../businesstime/user'
```
dheniges: I think we should stick with either businessUser or userBusiness. In the projects controller you have businessProject

dheniges (changes requested):
It all makes sense to me. I think you left an extraneous error in there, plus one naming nitpick

api/src/controllers/projects.js
jjidt: haha

dheniges: blarp is still here?

jjidt: haha blarp forever

dheniges (dismissed):
Blarp is dead, long live Blarp! 👍

jjidt:
@dheniges realized I forgot status codes and error condition tests, added those if you wanna take a quick glance at that latest commit

dheniges (approved):
Updated tests still pass 👍

PR #35, opened 2019-03-13
update test compose file so tests can run

dheniges:
- Combine build and execution into one step in makefile
- Remove `restart: always` from db container


jjidt (approved):
👍

PR #36, opened 2019-03-13
Project Form

jcontonio:
* Adds new ProjectFormComponent that can be used by CreatorContainer and future EditContainer
* Adds React-Select and a custom DropdownComponent that can take button and menu props for customization
* Going to reuse DropdownComponent anywhere there's a need for it (NavigatorContainer, GlobaSearch, etc)
* Adds VS Code resolvers
* Adds button and icon styles
* Organizes styles a little better, and adding some default layouts


dheniges:
:+1: sweet dropdown component. I think the change handlers in the container are perfect for the form pattern, nice!

dheniges (approved):
same thing I said in the comments, but in review form this time

PR #37, opened 2019-03-13
get users

jjidt:
- adds endpoint for fetching all users (with TODOs to convert to organization user fetch)
- user fetch bussinesstime integration tests
- creates files for db test setup & user factory
- now have text color in docker-compose test output
- user list fetching in the HeaderContainer -> Header component, with passing of first user value to header as currentUser (with TODO to change this to current user fetching)
- removes db command logging by default with env variable to turn on
- adds rebuilddev, seeddev, and unseeddev make commands


jcontonio (dismissed):
Can run make rebuilddev and make seedev, make start.... I see users 👍 
Only thing I can see is updating the README to link to docs/data.md, and update the initial setup instructions if we're replacing anything with make seeddev

api/src/businesstime/user.js
```diff
@@ -2,6 +2,7 @@ import { getDb } from '../db/db-connector'
 import ono from 'ono'
 
 const MODEL_NAME = 'User'
+const EXCLUDED_RESPONSE_FIELDS = ['password']
```
dheniges: I'm a fan of whitelists for return fields, vs. blacklisting something. Mostly it forces us to consider what we're returning from an API that will be public

api/src/db/db-connector.js
```diff
@@ -2,14 +2,16 @@ import Sequelize from 'sequelize'
 import path from 'path'
 import { modelFilePaths } from './models'
 
+const LOG_DB_COMMANDS = process.env.LOG_DB_COMMANDS === 'true'
```
dheniges, line 5: 👏

Makefile
```diff
@@ -10,6 +10,12 @@ rebuilddev:
 cleardev:
 	docker-compose down
 
+seeddev:
+	docker-compose run api ./run.sh seeddev --rm
+
+unseeddev:
```
dheniges: I exposed the sequelize command partially to avoid re-exposing each sequelize API via the Makefile via run.sh via package.json then via sequelize. Seems `make sequelize db:seed:all` is not much worse than `make seeddev`?

I didn't realize docker-compose run would start dependent services, which is nice. But this doesn't work if the container is already running. Whereas, currently, `make sequelize db:seed:all` won't work if the container is not running.

dheniges (dismissed):
I'd like to go for whitelisting return fields from the API, unless you feel strongly otherwise.

Also what do you think about the Makefile and sequelize? My preference there is to standardize on one approach, and then make sure the docs are accurate.

Makefile
jcontonio: My $.02 is I like the explicit `make sequelize db:seed:all`, and having one way of doing migrations/seeds vs having a bunch of make commands and keeping track of them.

jjidt: yeah agreed, I'll update that

jcontonio (approved):
Looks good 👍 thanks for adding to the project thing!

PR #38, opened 2019-03-14
updated prettier

dheniges:


dheniges:
So there's still an issue with husky hooks only running for one of the package's, so it won't run for both API and web. However, this config fixes eslint/prettier integrations somewhat 🤷‍♀️

jcontonio (approved):
🐳

PR #40, opened 2019-03-14
Dropdown component updates

jcontonio:
* Fixed the Prettier junk back to what we had
* Added max-len to eslintrc
* Added `hasAutoComplete` option to DropdownComponent so we can render a plain list of links if we want to, or whatever we want... 
* Fixed the bug where Header and NavigatorContainer weren't getting a project at all, because `match.params` was blank when `<Header />` was at the root level
* Changed NavigatorContainer to use DropdownComponent
* Moved the PATH_PROJECT_CREATE path constant somewhere else, so we didn't need Switch


web/src/client/js/hooks/useClickOutside.js
```diff
@@ -10,9 +10,18 @@ function useClickOutside(ref, callback) {
         callback()
       }
     }
+    function keyDownHandler(e) {
+      if (ref && ref.current) {
+        if (e.keyCode === 27) { // escape
```
jjidt: do we need a nested if here?

might be cleaner to do a single` if (ref && ref.current && e.keyCode === 27) {}`

web/src/client/js/sections/Dashboard/DashboardSection.js
```diff
@@ -1,21 +1,25 @@
 import React from 'react'
 import { Link } from 'react-router-dom'
 
-import ProjectsListContainer from 'Components/ProjectsList/ProjectsListContainer'
 import Constants from 'Shared/constants'
+import Header from 'Components/Header/HeaderContainer'
```
jjidt: lets use HeaderContainer here for consistency

web/src/client/js/sections/Project/ProjectNew.js
```diff
@@ -1,17 +1,21 @@
 import React from 'react'
 
+import Header from 'Components/Header/HeaderContainer'
```
jjidt: same

web/src/client/js/sections/Project/ProjectSection.js
```diff
@@ -1,15 +1,19 @@
 import React from 'react'
 
+import Header from 'Components/Header/HeaderContainer'
```
jjidt: same

web/src/client/js/sections/Projects/ProjectsSection.js
```diff
@@ -2,17 +2,23 @@ import React from 'react'
 import { Link } from 'react-router-dom'
 
 import Constants from 'Shared/constants'
+import Header from 'Components/Header/HeaderContainer'
```
jjidt: same

jjidt (dismissed):
works splendidly, just a couple comments

web/src/client/js/components/Dropdown/DropdownListComponent.js
```diff
@@ -0,0 +1,29 @@
...
+  onChange: PropTypes.func,
+  options: PropTypes.arrayOf(
+    PropTypes.shape({
+      value: PropTypes.string,
+      label: PropTypes.string
+    })
+  ).isRequired
+}
+
+export default NavigatorItemList
```
dheniges: The filename and component name aren't in sync. Looks like it should be `DropDownListComponent` now?

web/src/client/js/components/Dropdown/DropdownComponent.js
```diff
@@ -28,21 +38,33 @@ const DropdownComponent = ({ button, menu }) => {
...
+        }}>
         {hasIcon && <span className={`icon ${button.icon}`} />}
-        {button.label} {menu.value.length > 0 ? ` (${menu.value.length})` : ''}
+        {button.label} {menu.value && menu.value.length > 0 ? ` (${menu.value.length})` : ''}
       </button>
-      <div className="menu menu--dropdown" hidden={!expanded}>
+      <div className={`menu ${menu.hasAutoComplete ? 'menu--with-select' : ''}`} hidden={!expanded}>
+        {!menu.hasAutoComplete &&
+        <DropdownListComponent
+          basePath={Constants.PATH_PROJECT}
```
dheniges: assuming this is a generic component, shouldn't this basePath be passed as a prop by whatever is using it?

web/src/client/js/sections/Dashboard/DashboardSection.js
```diff
@@ -1,21 +1,25 @@
...
   render() {
     return (
-      <div role="main">
-        <div className="section">
-          <h1>FIRST NAME’s Projects</h1>
-          <Link to={Constants.PATH_PROJECTS} className="btn btn--blank">
-            All projects
-          </Link>
-          <ProjectsListContainer />
+      <>
```
dheniges, line 11: Is this the new `<Fragment>`?

dheniges (changes requested):
Something is weird with the buttons on the project create form. Mouse over them on and off, change the users, mouse over again. Sometimes it registers hover state and changes the cursor, other times it doesn't recognize it's a clickable button. No idea what's going on there

web/src/client/js/sections/Dashboard/DashboardSection.js
jcontonio: Yeah, @jjidt pointed that out yesterday

web/src/client/js/components/Dropdown/DropdownComponent.js
jcontonio: Oh good point... Actually, do you see a better way to do this? Maybe you pass it down as children so it renders _whatever_?

dheniges:
Not sure if this is a new change or not, but clicking on the navigator menu when it's open doesn't close the menu.

web/src/client/js/components/Dropdown/DropdownComponent.js
```diff
@@ -1,25 +1,28 @@
...
   const hasIcon = button.icon !== undefined
+  const selectRef = useRef()
   // Custom hooks
   const nodeRef = useRef()
   const collapseCallback = useCallback(() => {
     setExpanded(false)
   }, [])
+  const toggleCallback = useCallback(() => {
+    setExpanded(!expanded)
+  }, [expanded])
```
dheniges: so you can use the format `setExpanded(e => !e)` and expanded is no longer a dependency of your function, as you're just telling react "hey, take the current state and do this with it" without knowing/caring what the current state actually is. More details https://overreacted.io/a-complete-guide-to-useeffect/#making-effects-self-sufficient

web/src/client/js/components/DropdownWithSelect/DropdownSelectComponent.js
```diff
@@ -0,0 +1,97 @@
...
+  const selectRef = useRef()
+  // Custom hooks
+  const nodeRef = useRef()
+  const collapseCallback = useCallback(() => {
+    setExpanded(false)
+    selectRef.current.blur()
+  }, [])
+  const toggleCallback = useCallback(() => {
+    setExpanded(!expanded)
+    if (!expanded) {
```
dheniges, line 24: This is an odd one. You're essentially relying on the fact that this callback is invoked and the value of `expanded` _will_ change but has not yet changed. It works though 🤷‍♂️

dheniges (dismissed):
I dig the keyboard shortcut hook 👍 

A couple things you can backlog or ignore or whatever:
- In multi-select, hovering over the delete X icon that turns red, the cursor is a text editor line thingamajig, not the clicky hand
- I'm not sure if this is react-select's behavior, but the search field is also the selected items container seems weird:
Screen Shot 2019-03-15 at 2 52 53 PM

web/src/client/js/components/Dropdown/DropdownComponent.js
jcontonio: Oh that's cool

web/src/client/js/components/DropdownWithSelect/DropdownSelectComponent.js
jcontonio: Yeah, this drove me nuts... there's no callback to setExpanded... even if you throw a timeout of 10sec in here, expanded will never be what you'd expect.

dheniges (approved):

PR #41, opened 2019-03-15
add persistent db volume

jjidt:


docker-compose.yml
```diff
@@ -47,10 +47,6 @@ services:
       POSTGRES_USER: bonkey
       POSTGRES_PASSWORD: bong
     volumes:
-      - .database_data:/var/lib/postgresql/dev_db
+      - ./.postgres-data:/var/lib/postgresql/data
```
dheniges, line 50: @jjidt was this mounted at the wrong path in the container? dev_db vs data?

jjidt: yeah, the path I changed it to is the default

dheniges (approved):

PR #42, opened 2019-03-15
Production Chunks

jcontonio:
* Adds an AsyncComponent to dynamically import components, which are automatically chunked in webpack
* Splits vendor files more, and explicitly tells what files we need for vendor bundles in both dev and prod
* Adds gzip compression of static assets and minifying using terser in production build
* Adds BundleAnalyzerPlugin to analyze the bundles. Report is generated on `npm run build` and is placed in web/webpack-report.html
* Reduces bundles a ton


dheniges:
The compression, minification and whatnot all looks good.

I'm not sure about the bundle analyzer being part of the production build. It looks like the tool offers a CLI, so if we explicitly want the report we can generate it with that, without including it in the production build.

With the current setup, this report will get generated during the build process and hang out in the docker image that gets launched to prod. We can add a step to remove the file, but I'd rather just not generate it during a production build.

jcontonio:
Added gzip of static assets in production - I don't know how to test this

web/config/webpack.shared.js
```diff
@@ -52,10 +52,11 @@ const SharedConfig = {
   optimization: {
     splitChunks: {
       chunks: 'all',
+      minChunks: 1,
       cacheGroups: {
         vendor: {
           name: 'vendor',
-          test: /[\\/]node_modules[\\/]/,
+          test: /[\\/]node_modules[\\/](bulma|react|react-dom|react-select|react-redux|react-router-dom|react-stripe-elements|redux|redux-thunk)[\\/]/,
```
dheniges: this needs react-loadable too now. We'll soon want a better solution for building this vendor bundle. Won't webpack chunking on its own break up a larger bundle into chunks without specifying vendor given an entrypoint? Or are there optimizations from explicitly separating them?

docs/Web.md
```diff
@@ -5,7 +5,9 @@ The setup for Express and Webpack is quite complex. Here is an outline on what's
 
 **Bundles**
 
-* Each individual bundle needs an entry in the `entryPoints` object in webpack.config.js
+For a report on bundle size, run `npm run build` in /web and check `webpack-report.html`.
```
dheniges: this comment is outdated

dheniges (changes requested):
Update on doc, and then a question on vendor splitting

docs/Web.md
```diff
@@ -22,3 +40,4 @@ The setup for Express and Webpack is quite complex. Here is an outline on what's
 * Webpack creates a manifest file of hashed bundles using the `WebpackAssetsManifest` plugin, and puts it in the public directory.
 * Express ignores Webpack middleware all together in production, assigning the dynamically hashed bundles to `app.locals.bundles` via the manifest.
 * Hot reloading is completely ignored in production – see `entryFunction` and `entryPoints` in webpack.config.js
+* We are gzipping / minifying / and splitting code using react-loader
```
dheniges: react-loader is removed now, right?

web/package.json
```diff
@@ -44,6 +45,9 @@
       ".+\\.(css|styl|less|sass|scss)$": "jest-transform-css"
     }
   },
+  "eslintIgnore": [
```
dheniges: Can this not go in the `.eslintrc` file?

dheniges (changes requested):
Two things, I can update them if you want.

Also, the production build does not work with webpack/express. We'll have to debug more and I don't know how to test it without an integration test. The _build_ will work, but after logging in there's a blank page with no console errors. It's definitely missing some bundles or something.

web/package.json
dheniges: from eslint docs:
> ESLint always ignores files in /node_modules/*

So I'm just going to remove

dheniges (approved):

PR #43, opened 2019-03-18
Docker updates

dheniges:
Lots of Docker updates based around improving the dev experience and creating a better test/deploy setup for CI

- Expose `npm install` commands
- Separates dev and production Dockerfiles because we don't have a canonical repository name for them yet and can't build one image from a second one, so integrating with Docker repos and CI/CD is problematic
- Open up the entrypoint in each Docker container to run any shell script
- Cleanup web package.json to separate web server production required modules vs. production _build_ modules. Build has to be a dev dependency unfortunately, unless we create a separate package for just the build
- Move database volume to a named volume. It's cleaner than binding to a local directory since we're not doing anything with the database data locally (we can copy the data out of the named volume if we ever want to)

Backlog items to consider:
- custom CLI tool to manage the local environment/docker. A Makefile is not the appropriate tool, it's really for creating a build artifact, not interacting with an environment, managing state etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,3 @@ typings/
 
 # next.js build output
 .next
-
-# Dev database data
-.postgres-data/
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,19 @@ build:
 	./build.sh
 
 builddev:
-	./build.sh && docker-compose build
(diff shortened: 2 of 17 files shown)


jcontonio (approved):
👍

PR #44, opened 2019-03-18
codefresh pipeline yaml

dheniges:
Adds a yaml file to do builds and tests on Codefresh!

Runs API and web tests, and does production builds. It should trigger on any open PR. Theoretically even this one.


jcontonio (approved):
👍 Pipeline looks good

PR #45, opened 2019-03-18
Adjusting currentResource - Stops re-rendering of Header

jcontonio:


web/src/client/js/components/Project/ProjectContainer.js
```diff
@@ -3,18 +3,19 @@ import PropTypes from 'prop-types'
 import { withRouter } from 'react-router-dom'
 
 import useDataService from '../../hooks/useDataService'
-import dataMapper from '../../libs/dataMapper'
+// import dataMapper from '../../libs/dataMapper'
```
dheniges: hashtag deleteme

web/src/client/js/components/Header/Header.js
```diff
@@ -1,58 +1,33 @@
...
+    </div>
   )
 }
 
 Header.propTypes = {
-  match: PropTypes.object.isRequired,
-  currentUser: PropTypes.object
+  brand: PropTypes.object,
+  match: PropTypes.object,
+  history: PropTypes.object
```
dheniges: history is no longer a used props on this component

web/src/client/js/libs/currentResource.js
```diff
@@ -1,38 +1,43 @@
...
-        fetchAction: () => {},
-        getLoadingStatusFromState: () => {},
-        getDataFromState: () => {}
-      }
-    },
-    idName: null
+function returnNull() {
+  return {
+    fetchAction: () => {},
+    getLoadingStatusFromState: () => {},
```
dheniges:
```suggestion
    getLoadingStatusFromState: () => false,
```
The current implementation has a bug where once the resource is set, if it moves to an undefined resource the previous resource will never be unset. Returning false here tells useDataService the resource is loaded and will then set the appropriate (undefined in this case) data and the component using the hook will update accordingly

dheniges (changes requested):
A few small changes, and see my suggestion (I tested it locally, it works).

I don't love the reliance on the history object because it's mutable and there are no guarantees the component will update when it's mutated, or update how you'd expect it to. But it seems to work 🤷‍♀️

**Update**
After further testing, the location object exposed by react router is immutable, so we should replace reading the pathname with `location.pathname` to avoid future state bugs

web/src/client/js/libs/currentResource.js
jcontonio: Nice

web/src/client/js/components/Navigator/NavigatorContainer.js
```diff
@@ -12,9 +12,11 @@ import DropdownSelectComponent from 'Components/DropdownSelect/DropdownSelectCom
 
 import './Navigator.scss'
 
-const NavigatorContainer = ({ match, history }) => {
+const NavigatorContainer = ({ history }) => {
   const { data: projects } = useDataService(dataMapper.projects.list())
-  const { data: project } = useDataService(currentResource(match), [match])
+  const { data: project } = useDataService(
+    currentResource('project', history.location.pathname), [history.location.pathname]
```
dheniges: still have some history action going on here

dheniges (changes requested):
Navigator's still pulling history, otherwise looks good

dheniges (approved):

PR #46, opened 2019-03-19
Unified Production + Dev webpack stats processing

dheniges:
## Production + dev webpack stats processing
- Marries the dev and production parsing of webpack stats into a single plugin
- On dev, the parsed data is dynamically added to app.locals.bundles
- On production, the parsed data is written to a file during the build step, then imported by express when the web server starts

## Original Draft PR Info

Testing production build:
`API_PUBLIC_URL=http://localhost:3001 npm run build`
`API_URL=http://localhost:3001 API_PUBLIC_URL=http://localhost:3001 JWT_SECRET=jwtauthsecret NODE_ENV=production node dist/server/app.js`

Run API with docker compose
`docker-compose run --service-ports api ./wait-for.sh db:5432 -- ./run.sh devstart`

Other stuffs:
```
Notes:
# change web .env API_URL=http://host.docker.internal:3001

docker build -f web/Dockerfile-prod -t danger/web web/
docker run --env-file web/src/server/.env -it -p 3000:3000  danger/web /bin/sh

# In another shell:
docker-compose run --service-ports api ./wait-for.sh db:5432 -- ./run.sh devstart

# Attach to web prod container:
docker exec -it 18187cb4f423 /bin/sh

# Seed API if it's running in docker-compose run mode above
docker exec b0ba73dddf07 ./node_modules/.bin/sequelize db:seed:all

# Run webpack build in dev mode in ./web
npx webpack --env.development --config ./config/webpack.dev.js

# Run web server in dev mode with babel:
npx babel-node -r node_modules/dotenv/config ./src/server/app.js dotenv_config_path=./src/server/.env
```

### DIFF
--- a/docs/Web.md
+++ b/docs/Web.md
@@ -4,6 +4,7 @@
 The setup for Express and Webpack is quite complex. This document outlines what is going on with both the Express web server, and the Webpacked javascript bundles.
 
 * [Running a Bundle Analysis](#bundle-analysis)
+* [Running the client in production, locally](#production-local)
 
 ## Bundle Analysis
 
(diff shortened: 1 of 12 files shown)


jcontonio:
Things 

1. Once you get to `import DashboardContainer`, it breaks
1. Once you remove DropdownSelectComponent, and DropdownComponent from entryPoints.js, you will see the error: `TypeError: api/ is not a valid URL.` - this is from js/api/index.js
1. So then I changed that to be a real URL, and now everything renders

`const baseURL = new URL('api/', 'http://localhost:3001') // VAR_API_URL`

The question is, why are there no errors at all? And why does removing those two components from entryPoints then show the error?

I pushed these changes to the PR so you can see

jcontonio:
So right there in the webpack config:

> A dynamically loaded module is *not* an entry point.

Most likely, these components should just be left alone.

dheniges:
I think the VAR_API_URL is because that arg wasn't getting passed right on the command line, double checking it but it's set in the webpack shared config

jcontonio (approved):
Nice, hot reloading is working 👍

PR #47, opened 2019-03-19
Projects List

jcontonio:
Styles projects list according to Sketch file

* Adds animate option for turning animating on and off


web/src/client/js/components/ProjectsList/ProjectListComponent.js
```diff
@@ -0,0 +1,44 @@
...
+    setActiveProjectId(null)
+  }, [])
+  useClickOutside(nodeRef, collapseCallback)
+
+  function projectToggleHandler(itemRef) {
+    if (itemRef === null) {
+      setActiveProjectId(null)
+      return false
+    }
+    setActiveProjectId(itemRef.current.getAttribute('data-projectid'))
```
dheniges: instead of passing the ref and needing to read from the DOM to get the id, can't `ProjectsListItem` pass the project id to this function instead of the ref?

dheniges (changes requested):
I think you can drop refs in one case here, otherwise looks sweet

web/src/client/js/components/ProjectsList/ProjectListComponent.js
jcontonio: Nice catch, thanks!

dheniges (dismissed):

dheniges (approved):
It still builds

PR #48, opened 2019-03-20
Add organizations with relationships to user and project models

dheniges:
- Doesn't change any existing endpoints
- Run `make rebuilddev` or `make sequelize db:migrate:undo && make sequelize db:migrate && make sequelize db:seed:all`
- Sets up two orgs in the seed data, with two projects per org, and at least 2 users per org


jjidt (approved):
Excellent 👍

PR #49, opened 2019-03-20
project crud

jjidt:
API endpoints and client-side connection


api/src/businesstime/user.test-i.js
```diff
@@ -24,24 +24,24 @@ describe('UserBusiness', () => {
     it('should return sanitized users as POJOs', () => {
       for (const user of users) {
         expect(user.password).toBe(undefined)
-        expect(user).toBeInstanceOf(Object)
+        expect(user.constructor).toBe(Object)
```
dheniges, line 27: Is this a way to guarantee a POJO vs. a class instance?

api/src/controllers/projects.js
```diff
@@ -5,14 +5,25 @@ import ono from 'ono'
...
+    description: Joi.string()
   })
 })
 
 const projectsController = function(router) {
   router.post('/', validate(projectsPayloadSchema), addProject)
   router.get('/', getProjects)
   router.get('/:id', getProject)
+  router.put('/:id', validate(projectsPayloadSchema), replaceProject)
+  router.patch('/:id', validate(projectsPatchSchema), patchProject)
```
dheniges: Hmm, are we ever really supporting the difference between PUT and PATCH? I seem to recall something about PATCH support being questionable but that info may be outdated. The endpoints look identical except for the required name schema for PUT. I don't know what the RESTful rules state, but by definition we're never letting you replace an entire project. Thoughts?

api/src/db/__testSetup__/factories/project.js
```diff
@@ -0,0 +1,23 @@
+import faker from 'faker'
+import { getDb } from '../../dbConnector'
+
+const getProjectData = function(override = {}) {
+  const defaultProps = {
+    name: faker.random.word(),
```
dheniges, line 6: Thoughts on faker in tests? Theoretically we may generate a failure we can't reproduce. Although maybe it will force us to add better error logging, which sounds like a good thing :)

web/src/client/js/components/Project/ProjectContainer.js
```diff
@@ -10,11 +11,19 @@ const ProjectContainer = ({ location }) => {
...
 
 ProjectContainer.propTypes = {
-  location: PropTypes.object.isRequired
+  history: PropTypes.object.isRequired,
+  location: PropTypes.object.isRequired,
+  dispatch: PropTypes.func.isRequired
 }
 
-export default withRouter(ProjectContainer)
+export default connect()(withRouter(ProjectContainer))
```
dheniges: Why is this component connected? I don't see it using `dispatch` but may be missing shenanigans

web/src/client/js/components/ProjectForm/ProjectCreatorContainer.js
```diff
@@ -1,25 +1,56 @@
 import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 import { withRouter } from 'react-router-dom'
+import { connect } from 'react-redux'
```
dheniges: It might be worth just loading the store if all that's being used is the dispatch function. `Store.dispatch(action)` and keeps the DOM cleaner

web/src/client/js/api/index.js
```diff
@@ -21,4 +21,19 @@ async function fetch(resource) {
...
+  const res = await axiosInstance.put(resource, body)
+  return res.data
+}
+
+async function remove(resource) {
+  const res = await axiosInstance.delete(resource)
+  return res.data
+}
+
+export { fetch, add, update, remove }
```
dheniges, line 39: I like this 👍

web/src/client/js/components/ProjectsList/ProjectListComponent.js
```diff
@@ -38,7 +47,9 @@ function ProjectsListComponent({ projects }) {
...
 
 ProjectsListComponent.propTypes = {
-  projects: PropTypes.object.isRequired
+  projects: PropTypes.object.isRequired,
+  history: PropTypes.object.isRequired,
+  dispatch: PropTypes.func.isRequired
 }
 
-export default ProjectsListComponent
+export default connect()(withRouter(ProjectsListComponent))
```
dheniges: Same comment on dispatch - can it just import Store and dispatch?

dheniges (dismissed):
*commence comment barf*

api/src/businesstime/user.test-i.js
jjidt: yeah, we were getting false positives with the previous syntax

api/src/controllers/projects.js
jjidt: yeah, if we're allowing partial update of user defined fields, it should be a PATCH according to REST rules. Whereas with a PUT we'd be passing back the entire object for replacement(with or without db defined fields, which Joi will strip off for us). I've used it in the past, but I don't have super strong feelings about it. I can check out some API docs for other services to see how widely it's being used in the wild.

dheniges: I haven't run into it on other APIs, but I am more casual observer/user of APIs than hardcore REST API guru. I think I'm most interested in "is it useful?" and "do people know what it is?" if it's in a public API

api/src/db/__testSetup__/factories/project.js
jjidt: yeah, I think we could wind up with either a validation error or a unique field error, which in either case would mean we were doing something wrong with the testing, either not clearing out the db between tests, or not passing an acceptable override. But yeah I agree with the logging comment there, whatever it is we should hopefully be able to track it down.

web/src/client/js/components/Project/ProjectContainer.js
jjidt: ah yeah, I originally had a delete button in here, but nixed it when I merged in Jay's work. I'll remove it

web/src/client/js/components/ProjectForm/ProjectCreatorContainer.js
jjidt: I was reading something about connect being better for some reason, but we are using Store.dispatch in the useDataService lib, so maybe I'll change all of these to use that instead and see if any issues arise

dheniges: connect() is good for the mapStateToProps because it does automatic comparisons of changes and prevents unnecessary re-renders.

For dispatch, it's essentially a static method that just tells the store "this thing happened" so I am not aware of any additions connect() adds besides:
1) exposing dispatch
2) exposing helpers to bind dispatch to callbacks so you don't need `this.dispatchCallback.bind(this)`

api/src/controllers/projects.js
jjidt: I'm gonna remove the PATCH endpoint for now, it's easy enough to re-implement if we want partial updates for anything down the road

jcontonio (approved):
After running the following, I can create and delete projects:

```
make sequelize db:migrate:undo:all
make sequelize db:migrate
make sequelize db:seed:all
```

PR #50, opened 2019-03-21
Project view

jcontonio:
* Creates a new toolbar component for easy reuse
* Adds right/left positioning of menus in Dropdown and DropdownSelect
* Starts the Project component with styling
* Starts SVG as React component pattern for reusable, resizable, color changing SVGs
* Breaks out Filter and Sort components so ToolbarComponent doesn't get messy

Still more to do/organize, and didn't get to the changing name bit...but we're supposed to be able to rename the project by clicking on the Project name, on the Project view. That could be a marked `contenteditable` and onChange, trigger an update to Project endpoint.

Screen Shot 2019-03-20 at 7 49 42 PM


jcontonio:
Also, if this looks good to you and you want to tie into any of it, feel free to merge over the week

jjidt (approved):
👍

PR #53, opened 2019-03-25
initial perms middleware work with org id filtering

dheniges:
Adds multiple permissions middlewares, one for generating reusable CRUD permissions, and a lower level perms middleware that can request permission to do anything.
A permission request involves two objects:
- requestorInfo
- requestedAction

With those two pieces of information, we should be able to validate any type of future permission, coming from any requester (user, API token, other). We'll have to build out the actual checking of permissions, but it won't impact express middleware or route handlers.

the BusinessUser and BusinessProject now automatically filter by org id (and require an org id)


api/src/businesstime/project.js
```diff
@@ -9,14 +9,15 @@ async function create(body) {
   return createdProject.get({ plain: true })
 }
 
-async function findAll(id) {
+async function findAll(orgId) {
+  console.info({ orgId })
```
jjidt: leftover logging

api/src/businesstime/user.js
```diff
@@ -20,14 +20,18 @@ async function updateByEmail(email, body) {
 }
 
 // TODO convert to find all users within organization
```
jjidt: can remove this TODO now

api/src/controllers/loader.js
```diff
@@ -19,7 +20,7 @@ const UNAUTHENTICATED_CONTROLLERS = {
 }
 
 const loadControllers = (router, controllers, middleware) => {
-  Object.keys(controllers).forEach((path) => {
+  Object.keys(controllers).forEach(path => {
```
jjidt: why'd the linter have you remove the parens in this file, shouldn't they be there?

api/src/controllers/projects.js
```diff
@@ -57,6 +66,8 @@ const replaceProject = async function(req, res) {
   const { id } = req.params
   const { body } = req
 
+  // TODO: don't let users change orgId
```
jjidt, line 69: is this referring to what you did on line 54? or is it in reference to changing the orgId on the actual user object?

api/src/libs/middleware/reqPermissionsMiddleware.js
```diff
@@ -0,0 +1,53 @@
...
+  if (typeof actionBuilder !== 'function') {
+    throw new Error('actionBuilder arg must be a function')
+  }
+  return async (req, res, next) => {
+    try {
+      const hasPermission = await checkPermission(req.requestorInfo, actionBuilder(req))
+      if (hasPermission) {
+        next()
+      } else {
+        // TODO: convert to ono?
```
jjidt: I think the normal error here is fine, probably only need ono if it's a rethrow or need to pass additional status code info, in this case we're just triggering a 403 for everything

jjidt (changes requested):
This looks and works great, super easy to follow

Just a few cleanup comments

api/src/controllers/projects.js
dheniges: it's in reference to not being able to change an org id on the project, like what I did on the users. I backlogged an asana item to whitelist update fields, so I'll remove the TODO then

api/src/controllers/loader.js
dheniges: I don't know. I re-saved it now and it converted it back. I didn't manually make the change. Le sigh.

dheniges:
@jjidt ready for re-review (I re-requested it, but not sure how those notifications work)

jjidt (approved):

PR #54, opened 2019-03-25
org api

jjidt:
Adds a handful of things for signing up a new org/user

- organization BusinessTime :tm:
- custom password reset JWT auth with an a separate secret and a resetKey stored on the user (for initial password setting and reset in the future)
- hooks up sign-up views that rely on the token in the registration url(so we can easily switch this over to use an email for setting initial password)
- backfills some tests


api/src/businesstime/organization.test-i.js
```diff
@@ -0,0 +1,70 @@
...
+    })
+
+    it('should return the saved organization as a POJO', () => {
+      expect(organization).toEqual(expect.objectContaining(organizationBody))
+      expect(organization.constructor).toBe(Object)
+    })
+  })
+
+  describe('#updateById', () => {
+    describe('when the target document is found', () => {
```
dheniges: I changed this in my PR in another file, but I notice you have the text in these tests all referring to "document"

web/src/server/controllers/sign-up.js
```diff
@@ -1,14 +1,72 @@
...
+const API = new DangerAPI(process.env.API_URL)
+
 const SignUpController = function(router) {
   router.get('/processing', auth.jwtMiddleware, (req, res) => {
     res.render('user/processing', renderBundles(req, 'Processing', 'index'))
   })
 
-  router.get('/registration', auth.jwtMiddleware, (req, res) => {
-    res.render('user/registration', renderBundles(req, 'Registration', 'index'))
+  router.post('/registration', auth.jwtMiddleware, registerOrganization)
```
dheniges: How is this auth flow supposed to work? I don't have a token at this point when I'm trying to register, so this endpoint will 401 me

dheniges (changes requested):
Reviewed the codes and tests 👍 
- There's a failing test on codefresh, but locally they pass for me. Doing a rebuild to see what's up
- I'm not sure how the registration flow is supposed to work

dheniges:
Update on the tests: definitely getting intermittent failures on Codefresh and locally, although locally they seem to pass more often than not. It seems like a potential race condition where the cleanup is deleting records before the test finishes maybe?

web/src/server/controllers/sign-up.js
jjidt: That's not supposed to be an auth'd endpoint, I must have had a token and forgot to pull that out

dheniges (approved):

PR #55, opened 2019-03-27
add api endpoint logging

jjidt:
set to 'dev' mode for now, which is the colorful, easily readable setting

when we have a production environment, we'll probably want to set up some environment specific settings


dheniges (approved):
Mmm logging

PR #56, opened 2019-03-27
fix symbol operator and logging deprecation warnings from sequelize

jjidt:


dheniges (approved):
- It logs from the database!
- No sequelize deprecation warning

PR #57, opened 2019-03-27
Docs + Fields db schema + models

dheniges:
- Bonus BusinessDocument
- Full schema (theoretically) for documents and associated fields


api/src/businesstime/document.js
```diff
@@ -0,0 +1,47 @@
...
+
+  const updatedDocument = await document.update(body)
+  return updatedDocument.get({ plain: true })
+}
+
+async function deleteById(id, orgId) {
+  const db = getDb()
+  const document = await db.model(MODEL_NAME).findOne({ where: { id, orgId } })
+
+  if (!document) throw ono({ code: 404 }, `Cannot update, document not found with id: ${id}`)
```
jjidt: error message should maybe say delete, not update

api/src/db/migrations/20190327201522-create-fields.js
```diff
@@ -0,0 +1,45 @@
...
+      docId: {
+        type: Sequelize.INTEGER
+      },
+      versionId: {
+        type: Sequelize.INTEGER
+      },
+      fieldValueId: {
+        type: Sequelize.INTEGER
+      },
+      type: {
```
jjidt: should this be locked down to an enum array of possible numbers?

api/src/db/migrations/20190327203736-create-field-type-number-values.js
```diff
@@ -0,0 +1,30 @@
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('FieldTypeNumberValues', {
```
jjidt: does this also need a reference to the type number constant ?

api/src/db/models/fields.js
```diff
@@ -0,0 +1,16 @@
+
+module.exports = (sequelize, DataTypes) => {
+  const Fields = sequelize.define('Field', {
+    orgId: DataTypes.INTEGER,
+    name: DataTypes.STRING,
+    docId: DataTypes.INTEGER,
+    versionId: DataTypes.INTEGER,
+    fieldValueId: DataTypes.INTEGER,
+    type: DataTypes.INTEGER,
```
jjidt: do we need validation that the fieldValueId reference has the same type as what is defined here?

api/src/db/models/fieldtypenumbervalues.js
```diff
@@ -0,0 +1,11 @@
+
+module.exports = (sequelize, DataTypes) => {
+  const FieldTypeNumberValues = sequelize.define('FieldTypeNumberValue', {
+    orgId: DataTypes.INTEGER,
```
jjidt: on these type []values items, since these belong to a field, do we want to store the fieldId on here as well?

jjidt (changes requested):
just have a few questions below

also, I'm thinking all the fields, versions, values etc. should be singular - 
(field, version, value) in the model/migration definitions

api/src/businesstime/document.js
dheniges: yep, copied from projects, I'll update both

api/src/db/migrations/20190327201522-create-fields.js
dheniges: probably but I don't know what those numbers are yet. I don't know much about enums, can we alter the enums in future migrations without impacting existing data?

api/src/db/migrations/20190327203736-create-field-type-number-values.js
dheniges: I don't think so. It'd be repetitive, by definition this table can only contain one type. Unless we have two types that get stored in the same table.

api/src/db/models/fields.js
dheniges: we should validate that, but is that possible at the database layer?

api/src/db/migrations/20190327201522-create-fields.js
jjidt: read some more about this and it seems like that's a pain to update, so ignore that comment

api/src/db/models/fieldtypenumbervalues.js
dheniges: This is the question I was asking in slack earlier. The relation is on the fields table as `fieldValueId`. The only reason I'd want to store it twice is if it was a performance optimization for a particular query. Having it on the fields table ensures each field can only have one value. Validating that at the db level would be harder in reverse as values can live in one of many tables

api/src/db/models/fields.js
jjidt: probably not, and I guess if the fieldValue is created based off of this integer, it probably shouldn't matter

api/src/db/migrations/20190327203736-create-field-type-number-values.js
jjidt: 👍

api/src/db/models/fieldtypenumbervalues.js
jjidt: 👍

dheniges:
Updated names in migrations and models to be singular. I think I accidentally generated those as plural

jjidt (approved):
nice 👍

PR #58, opened 2019-03-27
remove email from jwt

jjidt:


dheniges (approved):
I was surprised my existing token worked with this, but since it had the userId in it I guess it makes sense. We're not validating if extra data's in the token.

Also confirmed it works on a new token/login

PR #59, opened 2019-03-27
dispatch correct project fetch action

dheniges:
This seems to solve the double fetch of a project, and the action is dispatched once. There may still be an underlying issue with useDataService dispatching multiple actions, I'm not positive it won't, but the solve is a bigger task. I have some stashed changes working on that if we need it.


jjidt (approved):
👍 nice catch

PR #60, opened 2019-03-28
Updating fix-entries webpack plugin

jcontonio:
Changes plugin to newest "stable" version


dheniges (approved):
Everything functions. What does this change/fix?

PR #61, opened 2019-03-28
easy seeding for field type values

dheniges:
- Adds constants for field types
- "documents" seeder coopted to seed docs with fields
- Easily add new docs with fields in db/seedData/docsWithFields


jjidt (approved):
Looks good 👍pushed up one change to assign constant ids to projects, and tie the the documents to a project

tested with my project/documents endpoint and seeing them come through, so if that all looks good to you, merge away

dheniges:
Hah, missed that critical piece, thanks 👍

PR #62, opened 2019-03-28
Log out user

jcontonio:
* Adds logout route, action, and reducer (in case we want to do anything in the future)
* Deletes token cookie
* Redirects to / using location.href from the reducer

### DIFF
--- a/web/src/client/js/components/Logout/LogoutComponent.js
+++ b/web/src/client/js/components/Logout/LogoutComponent.js
@@ -1,0 +1,9 @@
+import { Users } from '../../actions'
+import Store from '../../reducers'
+
+const LogoutComponent = () => {
+  Store.dispatch(Users.logout())
+  return null
+}
+
+export default LogoutComponent
--- a/web/src/client/js/components/Logout/LogoutComponent.js
+++ b/web/src/client/js/components/Logout/LogoutComponent.js
@@ -1,9 +1,0 @@
-import { Users } from '../../actions'
-import Store from '../../reducers'
-
-const LogoutComponent = () => {
-  Store.dispatch(Users.logout())
-  return null
-}
-
-export default LogoutComponent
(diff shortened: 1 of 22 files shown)


jcontonio:
I forgot, this also fixes buttons

web/src/client/js/components/Logout/LogoutComponent.js
```diff
@@ -0,0 +1,9 @@
+import { Users } from '../../actions'
+import Store from '../../reducers'
+
+const LogoutComponent = () => {
+  Store.dispatch(Users.logout())
```
dheniges: I think this is a nice spot for the `useEffect` hook with an empty dependency array so it only runs on mount. We only ever want this dispatch to occur once if this component is in the render path.

I like having the logout action, but I think the redirect logic and cookie setting should live in an action. So we'd dispatch a 'logout' action that dispatches to the store, sets the cookie and then does the redirect. It'd be similar to an action that fetches data from the API.

web/src/client/js/reducers/users.js
```diff
@@ -33,6 +33,11 @@ export const users = (state = initialState, action) => {
       const loadingById = { ...state.loading.byId, [id]: false }
       return { ...state, usersById, loading: { ...state.loading, byId: loadingById } }
     }
+    case ActionTypes.LOGOUT_USER: {
+      document.cookie = 'token=; expires=Thu, 01 Jan 1970 00:00:01 GMT; path=/'
+      location.href = '/'
```
dheniges: I think the store reducers should only ever respond to an action and data in the action to change state. See my comment on wrapping it in an action function

dheniges (changes requested):
It's completely functional, but I think the logic for logout should live in an action, not the store

dheniges (approved):

PR #63, opened 2019-03-28
Force validation of all fields to PUT/POST to whitelist update fields

dheniges:
- updates payloadValidation middleware to, by default, not allow unknown fields (with option to override if needed)
- payloadValidation now always validates req.body (can update to allow validating other stuff later if needed)
- orgId manually assigned in controller on project create or edit. Not sure if there's a better way to handle this? Since we're always forcing orgs, having the user send an orgId seems unnecessary, and saves us having to validate the orgId they sent is actually the orgId they belong to
- On web, commented out submission of team members to project create. I'm guessing that will likely be a separate endpoint, but it can submit through one form

### DIFF
--- a/api/src/controllers/organizations.js
+++ b/api/src/controllers/organizations.js
@@ -4,9 +4,7 @@ import BusinessOrganization from '../businesstime/organization'
 import ono from 'ono'
 
 const organizationsPayloadSchema = Joi.compile({
-  body: Joi.object().keys({
-    name: Joi.string().required()
-  })
+  name: Joi.string().required()
 })
 
 const organizationsController = function(router) {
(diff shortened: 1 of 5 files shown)


dheniges:
Looking at the Shopify API for inspiration, here's what they do:

## Shopify
- PUT endpoint will update whatever fields you pass, EXCEPT id. It looks like the id is always required in the object you're updating
- they show the GET endpoint returning createdAt/updatedAt, but not sure what happens if you PUT those fields. They do not PUT those fields in the examples

## Thoughts
- I like optional fields. If I don't PUT it, don't change it
- On the flip side, if I _do_ PUT it, it should change. For everything except ID. I like forcing all updates to be intentional.
- I don't think we should accept non-defined fields on an endpoint. If you send a field we don't recognize, as a user it's helpful to me to know that field is invalid. Otherwise if I am intentionally sending it, I would expect it to do something.
- Joi looks like it provides everything we need to return custom validation information: https://github.com/hapijs/joi/blob/v14.3.1/API.md#errors
- I don't think we want to blindly return all errors from Joi, but I think we could build a wrapper to whitelist the kinds of validation errors we want to return. Or somehow use Joi's validation to build the error response we want to send to the user

jjidt (approved):
I'm on board with everything you laid out here 👍

PR #64, opened 2019-03-29
Type / Icon cleanup

jcontonio:
* Made a bunch of small changes to type, and documented the pixel values
* Changed to redirect to projects on login
* Removes self from users list in project creation
* Complete test of icon as component
* Polishing up some spacing to the mockup


jjidt (commented):
this is looking good, only think I'm noticing is that if you throttle the browser on chrome, the filter and sortby icons show up in a column before popping over to the right, is that something you want to fix with this PR?

jjidt:
Sounds like it's mostly unrelated and we're gonna tackle the previous comment in a different PR

jjidt (approved):

PR #65, opened 2019-03-29
document endpoints

jjidt:
- CRUD endpoints for project documents
- tests
- adds param validation - (ie. we can validate http://localhost:3001/api/projects/blah/documents/33 and get a validation error for the non-number projectId param without having to do a db lookup)
- on validation, assign the validated value to req.body or req.param, which allows us to automatically cast to the defined type( I needed to compare ids and was getting false-negatives  with string vs number types )
- controller loader now lets us define child routes, so that we can access params from the parent route, in this case projectId


api/src/controllers/projectDocuments.js
```diff
@@ -0,0 +1,117 @@
...
+import crudPerms from '../libs/middleware/reqCrudPerms'
+import RESOURCE_TYPES from '../constants/resourceTypes'
+
+const { listPerm, readPerm, createPerm, deletePerm, updatePerm } = crudPerms(
+  RESOURCE_TYPES.PROJECT,
+  req => req.params.projectId
+)
+
+// TODO: this is currently piggybacking off of project perms,
+// does it need its own, or will it always be the same?
```
dheniges, line 14: I think for now it can use project perms, but I'd keep the comment/todo in here

dheniges (approved):
Nice work on this!
- child route loading 👍 
- validate body and params with the middleware 👍

PR #66, opened 2019-03-30
Documents UI

jcontonio:
* Adds start of very generic looking components for documents list and document
* Adds route for document (/projects/:projectId/document/:documentId)
* Have this here so we have something to load documents with, will make it all actually work once we have the API/Redux stuff in place

image


jcontonio:
Also, looking at this... I don't know that we even need the green project title with the icon at the top since it's already in the Navigator

I plan on adding a nice looking ui when no document is selected (/project/1)

dheniges (commented):
This seems like a good shell for being able to plugin document endpoints to. I like having clickable things in there, even when stuff is not functional. I'll talk with double J at checkin about how he wants to handle, whether to merge this PR or build on top of it.

dheniges (approved):
Looks good

PR #67, opened 2019-04-02
Project documents list in web

dheniges:
- Plugs in projectDocuments (which I think will be distinct from a Document, which will have fields)
- Separating projectDocs from Docs in redux will make managing state in two stores simpler, although we may repeat the base document fields in each store
- I don't know why my api package-lock.json file is changed. I redid an npm install with node v10.15.1, npm v6.4.1


jjidt (approved):
Looks/works great, we talked about adding an endpoint to get populated documents and possibly storing those separately, just throwing that here as a note

PR #68, opened 2019-04-02
project user assignment

dheniges:
@jjidt check out the BusinessProjectUser to see my lovely manual sql query. It would be nicer to use the sequelize associations, but I could not get them to work, even the more basic `belongsTo/hasMany` between the ProjectsUser table and Users.


jjidt (commented):
@dheniges might have to have you walk me through this, I'm not completely understanding why we need an intermediary ProjectUser model here. Would it work to establish the Project.hasMany(User) relationship and use the provided Project.addUser(*id*) and/or Project.setUsers([id, id, id]) functions?

dheniges:
We can add the full relationship from Projects -> Users (I left the commented belongsToMany code in place), and it may be helpful when a user fetches a list of projects to pull projects that way, User -> Projects. But as I was writing this I wound up not wanting/needing to do anything with a project object in most cases, and in many cases the sequelize methods weren't useful (eg removing a user from a project, I don't want to load all that user's projects, or all the users for the project. I just want to delete that join record). So this PR still utilizes sequelize joins and query generation, it just goes from the join table out to either resource while it's still setup to do the full traversal if we ever want. 

So since we're already dealing with assignment endpoints it's faster to work directly on the join table, both in terms of performance and writing queries. I think it also warrants its own BusinessTime layer because it'd clutter up projects (although conceptually, and based on the endpoints, it is a project function, not a user function).

api/src/db/models/user.js
```diff
@@ -15,6 +15,17 @@ export default function(sequelize, DataTypes) {
     User.belongsTo(models.Organization, {
       foreignKey: 'orgId'
     })
+    User.hasMany(models.ProjectUser, {
+      foreignKey: 'userId',
+      constraints: false,
+      // as: 'ProjectUser'
```
jjidt: can cleanup these comments, or TODO if you think we'll need any of this down the road

api/src/db/models/project.js
```diff
@@ -12,6 +12,11 @@ export default function(sequelize, DataTypes) {
     Project.belongsTo(models.Organization, {
       foreignKey: 'orgId'
     })
+    // Project.belongsToMany(models.User, {
```
jjidt: delete or TODO?

jjidt (dismissed):
looks great, really nice tests, just a couple comments that need deleted or TODO'd and then merge away

dheniges:
@jjidt removed comments, it dismissed your review

jjidt (approved):

PR #69, opened 2019-04-04
Documents list + new Project look

jcontonio:
* Adjusts a lot of icon bits to components for reuse
* Adjusting Dropdowns so they don't calculate their own opening positions
* New project view with documents list on the left, document on the right
* Start of a new Navigator look


dheniges (approved):
Looks good, although I can't actually review the css at all. Took me a minute to wrap my head around the debounce function, but that's cool.

PR #70, opened 2019-04-05
field endpoints

jjidt:
- document/field endpoints
- field instance methods for value setting/getting
- monitored database to ensure values were getting cascade deletion on field deletion


api/src/controllers/documentFields.js
```diff
@@ -0,0 +1,133 @@
...
+    res.status(e.code || 500).json({ error: `error fetching document with id ${id}` })
+  }
+}
+
+const addDocumentField = async function(req, res) {
+  const { body } = req
+  const { projectId, documentId } = req.params
+  const { orgId } = req.requestorInfo
+  // Overwrite orgId even if they passed anything in
+  body.orgId = orgId
```
dheniges, line 89: I recall this code somewhere else, eventually that should probably be an error message if they pass an invalid org id? This works for now though.

api/src/db/__testSetup__/initializeTestDb.js
```diff
@@ -8,7 +8,9 @@ afterAll(() => {
 export const clearDb = async function() {
   const db = getDb()
 
-  await db.truncate()
+  await Promise.all(Object.values(db.models).map((model) => {
+    return model.destroy({ truncate: true, cascade: true })
```
dheniges, line 12: What does this do that truncate() wasn't?

api/src/db/models/field.js
```diff
@@ -1,16 +1,59 @@
...
+  }
+
+  Field.prototype.setFieldValue = function(fieldValue) {
+    const modelName = fieldTypes.FIELD_TYPE_MODELS[this.type]
+    return this[`set${modelName}`](fieldValue)
+  }
+
+  Field.prototype.getFieldValue = function(fieldValue) {
+    const modelName = fieldTypes.FIELD_TYPE_MODELS[this.type]
+    return this[`get${modelName}`](fieldValue)
```
dheniges: isn't `fieldValue` here and two lines above always going to be undefined since this is a getter?

dheniges (changes requested):
JJ you've created a masterpiece. This is much more straightforward than I was anticipating, thanks to those pesky relationships seeming to do their job.

Looks like an unnecessary arg on a getter, otherwise it's solid

api/src/db/__testSetup__/initializeTestDb.js
jjidt: after defining the value/field relationship with the cascading delete, it started to throw an error message on truncate, this fixed it

api/src/db/models/field.js
jjidt: oops, yeah copypasta

api/src/controllers/documentFields.js
jjidt: yeah, we're doing this elsewhere, we could probably tackle all those in a PR

dheniges (approved):

PR #71, opened 2019-04-09
basic api docs

jjidt:
- api docs for project/document/field routes
- defined with openapi 3 spec
- recommend using the `Swagger Viewer` vsCode plugin to look at these locally
- I don't think we need to decide on a documentation template engine for this yet, can push that down the road a little, lots of options with this format, and openapi is easily converted to other specs if needed

lots of copy-paste happening in there, so that's probably something to look out for

### DIFF
--- a/api/openapi.yml
+++ b/api/openapi.yml
@@ -3,14 +3,10 @@ info:
   version: 1.0.0
   title: Project Danger API
   description: Public API spec for Project Danger
-  termsOfService: http://swagger.io/terms/
   contact:
     name: Project Danger Team
-    email: team@bonkeybong.com
+    email: info@bonkeybong.com
     url: http://www.bonkeybong.com
-  license:
-    name: Apache 2.0
-    url: https://www.apache.org/licenses/LICENSE-2.0.html
 servers:
   - url: http://localhost:3001
 paths:
--- a/api/openapi.yml
+++ b/api/openapi.yml
@@ -1,0 +1,574 @@
+openapi: "3.0.2"
+info:
+  version: 1.0.0
+  title: Project Danger API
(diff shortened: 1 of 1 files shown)


api/openapi.yml
```diff
@@ -0,0 +1,574 @@
...
+  title: Project Danger API
+  description: Public API spec for Project Danger
+  termsOfService: http://swagger.io/terms/
+  contact:
+    name: Project Danger Team
+    email: team@bonkeybong.com
+    url: http://www.bonkeybong.com
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
```
dheniges: Is this license for the API, or the docs? I'm thinking we don't want an Apache license on it either way

api/openapi.yml
```diff
@@ -0,0 +1,574 @@
+openapi: "3.0.2"
+info:
+  version: 1.0.0
+  title: Project Danger API
+  description: Public API spec for Project Danger
+  termsOfService: http://swagger.io/terms/
```
dheniges: Remove this ToS

api/openapi.yml
```diff
@@ -0,0 +1,574 @@
...
+  contact:
+    name: Project Danger Team
+    email: team@bonkeybong.com
```
dheniges: for now, `info@bonkeybong.com` since that actually exists

dheniges (changes requested):
This is cool. A few nitpick items in the metadata just so we don't forget about them. I didn't read through the docs line by line, but that swagger viewer is nice.

I'm assuming the spec would let you break up definitions across files somehow?

dheniges (approved):

PR #72, opened 2019-04-09
Navigator and Project Settings

jcontonio:
* New Navigator component and Menu styling
* Split routing up per "section" so that's a little cleaner
* Adding in ProjectSettings with ability to load settings panels based on route
* Added check in ProjectDocuments reducer for when a project has no documents, but not sure how to mark loading as false, since there is no ID
* Adds Tabs styling whenever we need them
* Adds new Form components for Textfield and Checkbox so we can keep everything consistent (more to come)


jcontonio:
This one will have the bug, my next PR doesn't... when testing this PR it's really only for styling / code style

web/src/client/js/hooks/useDetectInputMode.js
```diff
@@ -1,5 +1,18 @@
 import { useEffect } from 'react'
 
+// Test via a getter in the options object to see if the passive property is accessed
```
dheniges, line 5: I think it's useful to link to the code/docs where this came from. I was like "wtf is this" and then found the MDN discussion on it: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Safely_detecting_option_support

dheniges (changes requested):
Small doc link addition would be useful, see comment.

I don't think I understand what a section is supposed to be. For instance, in the Project section directory, some of them are only routing, some of them are just rendering a container, and some are rendering stuff + container(s).

dheniges:
Also this PR is 58 files, which is huge and hard to review. Each of the bullet points in your PR description could realistically be its own PR. If you combine one or two things that's cool, but this is too much for me to usefully review.

web/src/client/js/hooks/useDetectInputMode.js
jcontonio: Added

jcontonio:
Re: Sections - Trying to differentiate "pages" or "sections" that just render containers out of the /components directory, and not littering the /components directory with things that just render markup. 

* Sections should be dumb and render containers for that section
* Sections should contain the specific markup for that page, to organize the containers on the page
* Sections will be loaded dynamically using code splitting in the future
* The current ProjectsSection could probably use some reorganizing like the ProjectSection

dheniges (approved):

PR #73, opened 2019-04-09
Requesting and Adding single documents

jcontonio:
* Moves all document stuff to one reducer
* Tracks loading of documents, and document lists separately
* Adds ability to update document name
* Loads a document based on URL
* Fixes bug of loading projects with no documents


jcontonio:
There's a known react warning on here about prop types... it's fixed in the other PR

web/src/client/js/libs/currentResource.js
```diff
@@ -23,23 +23,48 @@ function returnNull() {
...
+      func = dataMapper.documents.id
+      if (matchResults) {
+        if (matchResults.params.projectId && matchResults.params.documentId) {
+          returnValue = func(matchResults.params.projectId, matchResults.params.documentId)
+        } else {
+          returnValue = returnNull()
+        }
+      } else {
+        returnValue = returnNull()
+      }
```
dheniges, line 60: This can be simplified:
```
if (matchResults && matchResults.params.projectId && matchResults.params.documentId) {
  returnValue = func(matchResults.params.projectId, matchResults.params.documentId)
} else {
  returnValue = returnNull()
}
```

web/src/client/js/reducers/documents.js
```diff
@@ -0,0 +1,94 @@
...
+        ...state,
+        loading: {
+          ...state.loading,
+          list
+        }
+      }
+    }
+    case ActionTypes.RECEIVE_DOCUMENTS: {
+      const loadingList = { ...state.loading.list, [action.projectId]: false }
+      if (action.data.length === 0) {
```
dheniges: Why is this returning early? Wouldn't we want the docs to get set to an empty array if they don't exist for a project?

dheniges (changes requested):
I like this PR :)  It is one feature

One question on an early return in the reducer for a project with no current docs, not sure why we wouldn't want to know there are no docs in the store?

web/src/client/js/reducers/documents.js
dheniges: Confirmed this if statement and its contents can safely be removed. The reduce function below will return an empty object if the document array (aka `action.data`) is empty here

dheniges (approved):

PR #75, opened 2019-04-10
fields redux flow

jjidt:
- sets up basic actions/reducer for field creation/update/deletion
- adds logging for validation errors on the API (we should eventually break out logging into its own lib so we can easily turn this off for production)


web/src/client/js/reducers/documentFields.js
```diff
@@ -0,0 +1,54 @@
...
+    case ActionTypes.INITIATE_FIELD_REMOVE: {
+      const { id } = action
+      const loadingById = { ...state.loading.byId, [id]: true }
+      return { ...state, loading: { ...state.loading, byId: loadingById } }
+    }
+    case ActionTypes.REMOVE_FIELD: {
+      const { id, docId } = action
+      const { documentFields } = this.state.documentFieldsById[docId] || {}
+      // eslint-disable-next-line no-unused-vars
+      const { [id]: __, ...restDocumentFields } = documentFields
```
dheniges, line 47: what is this madness?

dheniges (approved):

PR #76, opened 2019-04-10
New document creation

jcontonio:
* Hooks up creating new documents
* Adds new UI store to react to UI actions
* Tried doing it a few ways, this way seems the fastest and most intuitive


dheniges (approved):
This looks good and works nicely.

For a future ticket, probably want a way to trigger document creation without hitting enter. Maybe useClickOutside? Or a button to add it

PR #77, opened 2019-04-11
shorter field urls

jjidt:
- remove projects/:projectId requirement for field endpoints
- build out perms to check project ownership for a document(need to DRY up these checks with the project perm file when we do the actual perm flow for those)

### DIFF
--- a/api/src/controllers/documents.js
+++ b/api/src/controllers/documents.js
@@ -1,0 +1,3 @@
+const documentsController = function(router) {}
+
+export default documentsController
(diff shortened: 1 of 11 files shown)


api/src/controllers/documents.js
```diff
@@ -0,0 +1,3 @@
+const documentsController = function(router) {}
+
+export default documentsController
```
dheniges, line 4: Sweet controller. But what is this for? Is it required to get the child routes to load in the loader?

dheniges (commented):
Looks good to me. Just a question I think I know the answer to

dheniges (approved):

PR #78, opened 2019-04-11
Organization Roles and Permissions

dheniges:
- adds org role to user schema
- Signup flow assigns created user as org owner
- adds constant definitions of roles, with permissions based on resources and actions
- Adds a role check helper to see if roles contain a particular resource action permission
- Implements org role check that gives org role access priority

### DIFF
--- a/api/src/constants/roles.js
+++ b/api/src/constants/roles.js
@@ -1,6 +1,30 @@
+import actions from './actions'
+import resourceTypes from './resourceTypes'
 
 export const ORGANIZATION_ROLE_IDS = {
   OWNER: 1,
   ADMIN: 2,
   USER: 3
+}
+
+const rIds = ORGANIZATION_ROLE_IDS
+
+export const ORGANIZATION_ROLES = {
+  [rIds.OWNER]: {
+    [resourceTypes.PROJECT_USER]: {
+      [actions.UPDATE]: true,
+      [actions.DELETE]: true,
+      [actions.READ]: true,
+      [actions.LIST]: true
+    }
+  },
+  [rIds.ADMIN]: {
+    [resourceTypes.PROJECT_USER]: {
+      [actions.UPDATE]: true,
+      [actions.DELETE]: true,
+      [actions.READ]: true,
+      [actions.LIST]: true
+    }
+  },
+  [rIds.USER]: {}
 }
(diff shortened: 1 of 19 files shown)


api/src/constants/roles.js
```diff
@@ -0,0 +1,30 @@
+import actions from './actions'
+import resourceTypes from './resourceTypes'
+
+export const ORGANIZATION_ROLE_IDS = {
+  OWNER: 1,
```
jjidt, line 5: Question: are we backing ourselves into a corner by defining roles with sequential IDs? would it be possible in the future that we would want to insert roles between the current ones, or have roles with similar permission levels but varying perms? so would string constants, or IDs starting at 10, 20, 30 to leave room make sense?

api/src/constants/roles.js
```diff
@@ -0,0 +1,30 @@
...
+  ADMIN: 2,
+  USER: 3
+}
+
+const rIds = ORGANIZATION_ROLE_IDS
+
+export const ORGANIZATION_ROLES = {
```
jjidt, line 12: I like the organization of these, easy to follow

jjidt (commented):
I like how this is set up, just one question

api/src/constants/roles.js
dheniges: Good question. If roles don't inherit permissions and don't have any order, it won't matter. If we decide we want to make role order matter, then it would be a problem. In my head I was treating these essentially as a string constant, except it happens to be an integer because it's the cheapest to store in the db 🤷‍♀️  If we did want order to matter, we could define an order in the role definition (constants for now, could be db later)

api/src/libs/permissions/permissions.js
```diff
@@ -43,8 +44,18 @@ const resourceToHandler = {
  */
 export default async (requestorInfo, requestedAction) => {
   const resourceAccessHandler = resourceToHandler[requestedAction.resourceType]
-  // TODO: do this here? Maybe access handlers don't care?
-  requestorInfo.roles = getOrganizationRole(requestorInfo)
+  const hasOrgPermission = checkRolePermissions(
```
jjidt, line 47: nice, this makes more sense here

jjidt (approved):
Tested it out 👍 works great

PR #79, opened 2019-04-11
update api docs for new field routes

jjidt:
updates some API documentation

seems like the vscode viewer we're using doesn't fully support oneOf correctly which is a newish thing in the OpenApi spec, so some field-props are currently double defined


jcontonio (approved):
Looks good, thanks 👍

PR #80, opened 2019-04-11
Web – Delete documents

jcontonio:
* Adds delete action and ui
* Removes the separate list/documentsById in documents reducer, only using projectDocumentsById
* Changes babelrc targets to browsers, instead of Node. Was getting bizarre behavior with the destructuring in the reducer where documents wouldn't remove from state

### DIFF
--- a/web/src/client/js/actions/index.js
+++ b/web/src/client/js/actions/index.js
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
 import { makeActionCreator } from '../utilities/redux'
 
 export const ActionTypes = {
@@ -15,10 +16,12 @@ export const ActionTypes = {
   // Documents
   REQUEST_DOCUMENTS: 'REQUEST_DOCUMENTS',
   RECEIVE_DOCUMENTS: 'RECEIVE_DOCUMENTS',
-  CREATE_DOCUMENT: 'CREATE_DOCUMENT',
-  UPDATE_DOCUMENT: 'UPDATE_DOCUMENT',
+  INITIATE_CREATE_DOCUMENT: 'INITIATE_CREATE_DOCUMENT',
   RECEIVE_CREATED_DOCUMENT: 'RECEIVE_CREATED_DOCUMENT',
+  INITIATE_UPDATE_DOCUMENT: 'INITIATE_UPDATE_DOCUMENT',
   RECEIVE_UPDATED_DOCUMENT: 'RECEIVE_UPDATED_DOCUMENT',
+  INITIATE_DOCUMENT_REMOVE: 'INITIATE_DOCUMENT_REMOVE',
+  DOCUMENT_REMOVED: 'DOCUMENT_REMOVED',
   REQUEST_DOCUMENT: 'REQUEST_DOCUMENT',
   RECEIVE_DOCUMENT: 'RECEIVE_DOCUMENT',
   // Fields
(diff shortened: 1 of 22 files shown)


web/src/client/js/actions/index.js
```diff
@@ -15,10 +16,12 @@ export const ActionTypes = {
   // Documents
   REQUEST_DOCUMENTS: 'REQUEST_DOCUMENTS',
   RECEIVE_DOCUMENTS: 'RECEIVE_DOCUMENTS',
-  CREATE_DOCUMENT: 'CREATE_DOCUMENT',
-  UPDATE_DOCUMENT: 'UPDATE_DOCUMENT',
+  INITIATE_CREATE_DOCUMENT: 'INITIATE_CREATE_DOCUMENT',
```
jjidt: kinda nitpicky, but the rest of the actions in this file have the resource before the action

so these would follow `INITIATE_DOCUMENT_CREATE`

web/src/client/js/actions/index.js
```diff
@@ -15,10 +16,12 @@ export const ActionTypes = {
...
   RECEIVE_CREATED_DOCUMENT: 'RECEIVE_CREATED_DOCUMENT',
+  INITIATE_UPDATE_DOCUMENT: 'INITIATE_UPDATE_DOCUMENT',
   RECEIVE_UPDATED_DOCUMENT: 'RECEIVE_UPDATED_DOCUMENT',
+  INITIATE_DOCUMENT_REMOVE: 'INITIATE_DOCUMENT_REMOVE',
+  DOCUMENT_REMOVED: 'DOCUMENT_REMOVED',
```
jjidt: `REMOVE_DOCUMENT` for consistency

web/src/client/js/actions/index.js
```diff
@@ -15,10 +16,12 @@ export const ActionTypes = {
   // Documents
   REQUEST_DOCUMENTS: 'REQUEST_DOCUMENTS',
   RECEIVE_DOCUMENTS: 'RECEIVE_DOCUMENTS',
-  CREATE_DOCUMENT: 'CREATE_DOCUMENT',
-  UPDATE_DOCUMENT: 'UPDATE_DOCUMENT',
+  INITIATE_CREATE_DOCUMENT: 'INITIATE_CREATE_DOCUMENT',
   RECEIVE_CREATED_DOCUMENT: 'RECEIVE_CREATED_DOCUMENT',
+  INITIATE_UPDATE_DOCUMENT: 'INITIATE_UPDATE_DOCUMENT',
```
jjidt: same

jjidt (dismissed):
Everything is looking really nice, just the one thing on the action names

dheniges (changes requested):
There's an issue with document update now, if you change the doc title the reducer throws an error when trying to update the doc.

On a side note, I think we need two babel configs, one for the web server, and one for the web client / webpack build. Since chrome is a target, I assume v8 should handle whatever chrome output and it seems to work. But it seems like we should be targeting node for the server. I'll open a separate issue for this.

web/src/client/js/components/Document/DocumentComponent.js
```diff
@@ -1,26 +1,54 @@
...
+  }
+  const changeHandlerAction = debounce(500, (e) => {
+    nameChangeHandler(e)
+  })
   return (
     <div className="document" key={docKey}>
       <header>
-        <textarea
+        <div
+          contentEditable
```
dheniges: according to MDN docs, this should explicitly be set to "true" https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable

dheniges: Also, is contenteditable the right tool for this? It's going to insert html entities and elements. If you put a few spaces in it inserts `&nbsp;`, which we don't want. The return key is overridden on this element, but if it's not hitting return inserts div elements with contenteditable.

dheniges (changes requested):
questions on contenteditable

web/src/client/js/components/Document/DocumentComponent.js
jcontonio: Re: contenteditable -> I had previously used textarea, which worked nicely, except that you can't style a textarea to line up with anything. Let me try some more things and see if I can get back to a textarea.

Note that documentList is also using contenteditable.

dheniges (approved):

PR #81, opened 2019-04-12
standardize api responses

jjidt:
- updates api responses as { data: responseData } on all endpoints
- updates client code to handle new format
- updates documentation

not sure if errors should be as they currently are with the message on the { error: '' } property of the response body,

or if we should do { message: '' }
or { error: { message: '' } }

### DIFF
--- a/web/src/client/js/api/index.js
+++ b/web/src/client/js/api/index.js
@@ -17,23 +17,23 @@ const axiosInstance = axios.create({
 })
 
 async function fetch(resource) {
-  const res = await axiosInstance.get(resource)
-  return res.data
+  const { data } = await axiosInstance.get(resource)
+  return data
 }
 
 async function add(resource, body) {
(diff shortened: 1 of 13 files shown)


web/src/client/js/api/index.js
```diff
@@ -17,23 +17,23 @@ const axiosInstance = axios.create({
 })
 
 async function fetch(resource) {
-  const res = await axiosInstance.get(resource)
-  return res.data
+  const { data } = await axiosInstance.get(resource)
+  return data
```
dheniges, line 21: oh that's sweet these are the only updates needed on the client

dheniges (approved):
I'll be honest, I thought this would be more work. It looks good to me, seems to work nicely. Re: errors, I think the current setup is fine until we want to flesh that out more and probably return specific error codes particular to the app or somesuch. I assume errors will become an object with more info

PR #82, opened 2019-04-12
Confirmation Component

jcontonio:
* Has work from Delete Document PR
* Adds UI state for rendering a confirmation dialog with cancel and confirm actions
* Remove max-len from eslintrc and any ignores we had in the web directory (sorry @dheniges)


dheniges (approved):
This is sweet 👍

PR #83, opened 2019-04-12
Project Roles

dheniges:
# Project Roles!
Integrates the permissions check to look at both org roles and project roles

### Permissions Concepts
1. Permissions determine if you have the org or project access based on role
1. Permissions **do not** determine if you have fine-grained access control in special cases
    eg an org admin cannot delete an org owner. This will need to be a special check on an endpoint
1. Org access is the first check, project access is the second check. So if an org role has a permission, it overrides any project permission, or lack thereof

### Default project access
Until we implement project access levels, this gives all org users project read access. the `projectPermissionGenerator` is setup to quickly integrate that feature when it goes live, with minimal code changes

### Introducing debug module
- Because console.info's are a helluva thing, but debugging is useful, enter `debug`
- It's a simple module that exposes a function that returns a function to log in a given namespace. It's widely used (add the `DEBUG=*` to your node api .env to see), and does nothing if not enabled. It's the cost of calling an empty function.
- You can namespace to your heart's desire. This PR introduces the project namespace `danger`, and exposes a permissions logger. So if you want to debug permissions in PD, set `DEBUG=danger:permissions`. Want all PD? `DEBUG=danger:*`


api/src/constants/roles.js
```diff
@@ -7,24 +7,128 @@ export const ORGANIZATION_ROLE_IDS = {
...
+  }
+}
+
+// All project roles are permissions _for an individual project_,
+// Project Role permissions only apply to
+// a user explicitly assigned the role on an individual project!
+export const PROJECT_ROLES = {
+  [PROJECT_ROLE_IDS.ADMIN]: {
+    [resourceTypes.PROJECT]: {
+      [actions.READ]: true,
```
jjidt, line 84: purposefully no list perm here right?

api/src/controllers/documentFields.js
```diff
@@ -9,7 +9,7 @@ import fieldTypes from '../constants/fieldTypes'
 
 const { listPerm, readPerm, createPerm, deletePerm, updatePerm } = crudPerms(
   RESOURCE_TYPES.DOCUMENT,
-  req => req.params.documentId
+  (req) => { return { documentId: req.params.documentId } }
```
jjidt, line 12: oh nice, this expects an object return now, i like it

jjidt (approved):
a lot in here like you said :) tests are solid, org perms are solid, tested as much as I could so gonna go ahead, this feels safe to merge and will get a little more manual testing as we implement this on specific endpoints

api/src/constants/roles.js
dheniges: yep

api/src/controllers/documentFields.js
dheniges: yeah, letting it define whatever data object it wants to now, much more flexible

PR #84, opened 2019-04-12
Adds useClickOutside to document creation

jcontonio:
* Adds a new route to create a document, so that we unload the current document and prevent a flash of document name
* This will allow us to go back to what I had before if we want... creating a new document has the user name the document inside the "document" (like Bear)
* Adds preventEscapeFunctionality to useClickOutside if we want escape to do something else
* Allows the user to give the document a name, click outside, and the doc gets made

### DIFF
--- a/web/src/client/js/components/DocumentsList/DocumentsListComponent.js
+++ b/web/src/client/js/components/DocumentsList/DocumentsListComponent.js
@@ -9,6 +9,7 @@ import DocumentsListItem from './DocumentsListItem'
 import './DocumentsList.scss'
 
 const DocumentsListComponent = ({ createChangeHandler, creating, createCallback, documents }) => {
+  const listItemRef = useRef()
   const nameRef = useRef()
 
   // Select the contents of the contentEditable div (new document name)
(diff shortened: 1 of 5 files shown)


web/src/client/js/components/DocumentsList/DocumentsListComponent.js
```diff
@@ -26,19 +28,36 @@ const DocumentsListComponent = ({ createChangeHandler, creating, createCallback,
     title: 'Create a new document in this project'
   }
 
+  function createDocument() {
+    if (nameRef.current.value !== 'New Document') {
```
jjidt: this is used in multiple places, so we should define 'New Document' as a constant

jjidt (dismissed):
It works! One small thing

web/src/client/js/components/DocumentsList/DocumentsListComponent.js
jcontonio: I left out the New Document button, because we may want to have a different defaultValue than the button language in the future.

dheniges (approved):
Works well 👍 

In the future we may want/need a different way to differentiate between a doc that hasn't been created and a real doc besides looking at the name "New Document". Because right now I can't create a doc titled "New Document"

PR #86, opened 2019-04-12
org role update and tests

jjidt:
adds a use org-role update controller,

currently no perms check on this, what do we want to do here?


dheniges (commented):
Thinking through perms, I think if a user has the `USER:CREATE` permission (which doesn't exist yet) that assumes they can assign the org role. I think we can then add a separate `UPDATE_ORG_ROLE` permission to the user resource type for this PR.

jjidt:
this is updated with a org perms check on the route, we'll need to address the problem of storing the orgRoleId in the JWT though

dheniges (approved):
v. nice

re: updated org role id. We do need a solve for that, but it's tangential to this work so 👍

PR #87, opened 2019-04-13
Fields creation

jcontonio:
* Adds UI creation and type action/reducer so different parts of the app can react to that (disable things, fade out, etc)
* Creates a name for the field automatically, and renders it on the page
* Ability to rename fields


web/src/client/js/components/DocumentFields/DocumentFieldsContainer.js
```diff
@@ -0,0 +1,53 @@
...
+DocumentFieldsContainer.propTypes = {
+  match: PropTypes.object.isRequired,
+  removeField: PropTypes.func.isRequired,
+  updateField: PropTypes.func.isRequired,
+  ui: PropTypes.object.isRequired,
+}
+
+const mapStateToProps = (state) => {
+  return {
+    ui: state.ui
```
dheniges: the only property used is `ui.fields.creating`, might as well just grab that here?

web/src/client/js/components/DocumentFields/DocumentFieldsContainer.js
```diff
@@ -0,0 +1,53 @@
...
+
+import useDataService from 'Hooks/useDataService'
+import dataMapper from 'Libs/dataMapper'
+import { updateField, removeField } from 'Actions/field'
+
+import DocumentFieldsComponent from './DocumentFieldsComponent'
+
+const DocumentFieldsContainer = ({ match, removeField, updateField, ui }) => {
+  const { documentId } = match.params
+  const uiState = ui.fields
```
dheniges: the `uiState` name is confusing because it's actually the field state. Looks like you can avoid this variable assignment, see comment on mapStateToProps in this component

web/src/shared/constants.js
```diff
@@ -16,6 +16,11 @@ module.exports = {
   PATH_PROJECT_CREATE: '/project/create',
   PATH_SETTINGS: '/settings',
   MAX_COOKIE_AGE_MS: '604800000', // 7 days
+  FIELD_LABELS: {
+    1: 'text-field-',
```
dheniges: as we add more field types, probably best to define these as `[FIELD_TYPES.STRING]: 'text-field-`

web/src/client/js/actions/index.js
```diff
@@ -39,6 +39,8 @@ export const ActionTypes = {
   LOGOUT_USER: 'LOGOUT_USER',
   // UI
   UI_DOCUMENT_CREATE: 'UI_DOCUMENT_CREATE',
+  UI_FIELD_CREATE: 'UI_FIELD_CREATE',
```
dheniges: I think for consistency we should have an `INITIATE_FIELD_CREATE` action that's not UI-specific instead of this. We use that pattern for update/delete actions. You can see the document actions follow this pattern as well. The UI store can read that action and respond the same way it does to this UI_FIELD_CREATE action.

web/src/client/js/components/DocumentFields/DocumentFieldsComponent.js
```diff
@@ -0,0 +1,57 @@
...
+  function renderFields() {
+    const fieldArray = []
+    const showFieldName = fields.length > 1
+    fields.forEach((field) => {
+      switch (field.type) {
+        case Constants.FIELD_TYPES.TEXT: {
+          return fieldArray.push(
+            <DocumentFieldComponent key={field.id} type={field.type} onDestroy={() => { fieldDestroyHandler(field.id) }}>
+              <FieldTextComponent field={field} showName={showFieldName} onChange={fieldChangeHandler} />
+            </DocumentFieldComponent>
```
dheniges: each of these case statements is doing the same thing, I think this could be DRY'd up in a function that returns the `FieldTypeComponent` inside the `DocumentFieldComponent` since the document field component wrapper is always the same

web/src/client/js/components/FieldNumber/FieldNumberComponent.js
```diff
@@ -0,0 +1,46 @@
...
+  // field name change
+  const nameRef = useRef()
+  const debouncedNameChangeHandler = debounce(1000, (e) => {
+    onChange(field.id, { name: nameRef.current.value })
+  })
+  // Value change
+  const valueRef = useRef()
+  const debouncedValueChangeHandler = debounce(1000, () => {
+    onChange(field.id, { value: valueRef.current.value })
+  })
```
dheniges: Same situation on these name change handlers and value change handlers. It looks like the same code across 3 files, so maybe some of it could be broken up into a component that handles the name display and change. Or a parent component that handles the name field and updates, and renders a child that is specific to the field type.

dheniges (changes requested):
Good stuffs here, fields bring joy to my heart

Feedback points:
- consistent action patterns (initiate create, receive create pattern)
- DRY code

dheniges (approved):

PR #88, opened 2019-04-15
updated update schemas

jjidt:
- allows empty field values on creation
- breaks out create and update schemas using helpers
- partialFields schema helper
- updates documentation

NOTE: the previous swagger viewer I suggested wasn't showing payload examples, use `OpenAPI Preview` extension instead

### DIFF
--- a/api/src/controllers/payloadSchemas/helpers.js
+++ b/api/src/controllers/payloadSchemas/helpers.js
@@ -1,8 +1,10 @@
 import Joi from 'joi'
 import * as user from './user'
+import * as project from './project'
 
 const resourceToSchema = {
-  user
+  user,
+  project
 }
 
 /**
--- a/api/src/controllers/payloadSchemas/helpers.js
+++ b/api/src/controllers/payloadSchemas/helpers.js
@@ -1,10 +1,14 @@
 import Joi from 'joi'
 import * as user from './user'
 import * as project from './project'
+import * as field from './field'
+import * as document from './document'
 
 const resourceToSchema = {
   user,
-  project
+  project,
+  field,
+  document
 }
 
 /**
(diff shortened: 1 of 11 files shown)


jcontonio (dismissed):
* I can update fields with just updating a name, or just updating a value 👍 
* I can create a field with just a name and a type (string or number) 👍 
* I get "name" is required if I try to create a field with just a value, that seems good, although I see details response is an array and I am only getting one validation error at a time. Any way to get all the errors?

Documentation updates:

* StringField, TextField, and NumberField all have value as required
* UpdateField only has `value`, but I'd expect `name` here as well so I can update the name
* Field only has `id` as required, expect `id` and `name` - name isn't listed in Field at all
* Should Document have `projectId` as required?

jcontonio:
Oh, also `order` on UpdateField?

jjidt:
update, I think the documentation correctly reflects the current state of the API

added order as a field return value, but we're not currently updating order, I'm thinking that will be it's own endpoint with side effects

the multi-validation error thing is probably a Joi setting, but we can handle that in the future

api/src/controllers/payloadSchemas/helpers.js
```diff
@@ -13,25 +19,53 @@ const resourceToSchema = {
...
 }
 
+/**
+ * Use the baseSchema of a resource and limit the fields for validation
+ * be required
+ * @param {String} resource the resource name
+ * @param  {...String} fields field names found in schema to use in validation
+ * @return {Joi Schema} schema the compiled schema with included fields
+ */
+const partialFields = (resource, ...fields) => {
```
dheniges, line 43: ah nice

dheniges (approved):
Looks good to me

PR #89, opened 2019-04-15
projectUser assignment endpoints

dheniges:
So I nuked the idea of these endpoints returning user objects now that there's data associated with the join (that data being the project role assigned currently, and maybe more in the future).

If we do want to get project users, we can add new endpoints at the old url of /projects/:id/users and/or support the include? syntax discussed on an endpoint.

I also blew up the husky hook because I'm the only one that runs it on web and it blew up tons of files and rewrote Jay's css which made him very happy


api/src/businesstime/projectuser.test-i.js
```diff
@@ -3,7 +3,9 @@ import ProjectFactory from '../db/__testSetup__/factories/project'
 import UserFactory from '../db/__testSetup__/factories/user'
 import initializeTestDb, { clearDb } from '../db/__testSetup__/initializeTestDb'
 import constants from '../db/__testSetup__/constants'
-import { PUBLIC_FIELDS as USER_PUBLIC_FIELDS } from './user'
+
+// TODO: in BusinessTime, define and filter by public fields somewhere
```
jjidt: is this something we need to wait on?

api/src/controllers/projectUsers.js
```diff
@@ -72,39 +72,39 @@ const getProjectUser = async function(req, res) {
...
   try {
-    await Promise.all(
-      body.users.map(u => BusinessProjectUser.deleteByUserIdForProject(u.id, projectId, orgId))
+    const projectUser = await BusinessProjectUser.updateProjectUserById(
+      id,
+      body.roleId,
+      orgId
     )
-    res.status(204).send()
+    res.status(201).json({ data: projectUser })
```
jjidt: should this be a 200 code on the update?

jjidt (commented):
Just a couple questions, but overall this seems like a good structure 💯

dheniges:
this is ready for re-review

jjidt (approved):
👍

PR #90, opened 2019-04-16
Field deletion

jcontonio:
* Deletes fields with confirmation


dheniges (dismissed):
👏 it's beautiful

dheniges (approved):
changed my mind, he's alright

PR #91, opened 2019-04-17
project - user  seed data

jjidt:


dheniges (approved):
Noice.

Project delete works fine for me. You might need to do a rebuild of dev or drop the db and migrate.

PR #92, opened 2019-04-17
Refactor BusinessTimey™ layers to remove orgId

dheniges:
- Define public fields by resource type in constants
- BusinessTime returns the public fields from all functions, unless it's special like user with `sanitized` function returns

### DIFF
--- a/api/src/businesstime/document.test-i.js
+++ b/api/src/businesstime/document.test-i.js
@@ -18,7 +18,7 @@ describe('BusinessDocument', () => {
   describe('#createForProject', () => {
     const documentBody = {
       name: 'test-document',
-      orgId: constants.ORG_ID,
+      orgId: constants.ORG_ID
     }
     let document
 
--- a/api/src/businesstime/document.test-i.js
+++ b/api/src/businesstime/document.test-i.js
@@ -16,21 +16,21 @@ describe('BusinessDocument', () => {
   describe('#createForProject', () => {
     const documentBody = {
       name: 'test-document',
-      orgId: constants.ORG_ID
+      orgId: constants.ORG_ID,
     }
     let document
 
     beforeAll(async () => {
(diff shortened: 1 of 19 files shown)


api/src/businesstime/document.test-i.js
```diff
@@ -16,21 +18,21 @@ describe('BusinessDocument', () => {
   describe('#createForProject', () => {
     const documentBody = {
       name: 'test-document',
-      orgId: constants.ORG_ID
+      orgId: constants.ORG_ID,
```
jjidt: trailing comma ?

api/src/libs/utils.js
```diff
@@ -0,0 +1,7 @@
+
+export const pick = (obj, fields) => {
```
jjidt, line 2: excellent 💯

api/src/constants/resourcePublicFields.js
```diff
@@ -0,0 +1,30 @@
...
+  'order'
+].concat(globalPublicFields)
+
+const PROJECT_USERS_PUBLIC_FIELDS = ['id', 'roleId', 'projectId', 'userId'].concat(globalPublicFields)
+
+const USER_PUBLIC_FIELDS = ['firstName', 'lastName', 'email', 'orgRoleId'].concat(globalPublicFields)
+
+
+export default {
+  [resourceTypes.DOCUMENT]: DOCUMENT_PUBLIC_FIELDS,
```
jjidt: super bonus option: we could export these all as [resourceTypes.DOCUMENT]: Object.freeze(DOCUMENT_PUBLIC_FIELDS)

to make sure nothing accidentally mutates the contents

jjidt (changes requested):
love the refactor, one small thing and one super bonus option

jjidt (approved):
👍 👍 👍 👍 👍 👍 👍 👍 👍

PR #93, opened 2019-04-17
field validation

jjidt:
- add option to include details about field validation errors (turned on for proj, doc, field)
- validate field values against type

### DIFF
--- a/api/src/businesstime/field.js
+++ b/api/src/businesstime/field.js
@@ -21,6 +21,8 @@ async function createForDocument(docId, body) {
   const db = getDb()
   const { orgId } = body
 
+  validateFieldValueByType(body.value, body.type)
+
   const document = await db.model('Document').findOne({ where: { id: docId, orgId } })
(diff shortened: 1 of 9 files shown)


api/src/businesstime/field.js
```diff
@@ -91,13 +96,35 @@ async function deleteByIdForDocument(id, docId, orgId) {
...
 
+function validateFieldValueByType(fieldValue, type) {
+  let isValid = false
+
+  // we're allowing null or undefined values for all types
+  if (fieldValue == null) return
+
+  switch (type) {
+    case 1:
+    case 2:
```
dheniges: Do we have constants for these types yet? Prolly should if we don't

dheniges (changes requested):
constants thing

dheniges:
endpoint returns and validation seem solid 👍

dheniges (approved):
yay

PR #94, opened 2019-04-17
Field styling and reordering

jcontonio:
* Styles document + fields
* Hides field tools (delete, so far) until hovered or focused
* Changes field name length to 50 MAX in API and in UI
* Does re-ordering of fields on drag and drop (UI only)


web/src/client/js/components/DocumentFields/DocumentFieldsContainer.js
```diff
@@ -29,22 +44,62 @@ const DocumentFieldsContainer = ({ creating, match, removeField, updateField, ui
...
+  function dropHandler(e) {
+    if (e.stopPropagation) {
+      e.stopPropagation()
+    }
+    const from = Number(draggingElement.dataset.order)
+    const to = Number(e.currentTarget.dataset.order)
+    e.currentTarget.classList.remove('field--dragging', 'field--dragged-over')
+    if (to === from) { return }
+    const fieldData = orderedFields
+    fieldData.splice(to, 0, fieldData.splice(from, 1)[0])
```
dheniges, line 80: I haven't confirmed, but the missing re-render issue may be in part because this is modifying the state `orderedFields` without calling setState (well, `setOrderedFields` in this case). My hunch is `setOrderedFields` is seeing that `orderedFields` is unchanged, because you've already modified it, thus preventing re-render

dheniges (changes requested):
I think I found the re-render bug

dheniges (approved):

PR #95, opened 2019-04-18
ProjectToken

dheniges:
- BusinesProjectToken + tests
- ProjectToken db model/migration

@jjidt  thinking next step is a coordinator layer to generate secret and token, then endpoints. Is that how you'd do this?


jjidt (approved):
most excellent 🤘

PR #96, opened 2019-04-18
Login redirect and flash

jcontonio:
* Redirects on bad login
* Adds a flash message using req.query

### DIFF
--- a/web/src/server/libs/auth.js
+++ b/web/src/server/libs/auth.js
@@ -14,7 +14,12 @@ passport.deserializeUser((id, done) => {
 localAuth(passport)
 JWTAuth(passport)
 
+const passportResponse = {
+  session: false,
+  failureRedirect: '/sign-in?message=login'
+}
+
 export default {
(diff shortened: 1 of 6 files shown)


web/src/server/libs/auth.js
```diff
@@ -14,7 +14,12 @@ passport.deserializeUser((id, done) => {
 localAuth(passport)
 JWTAuth(passport)
 
+const passportResponse = {
+  session: false,
+  failureRedirect: '/sign-in?message=login'
```
dheniges: setting `failureRedirect` here causes the following problem: if I logout and then go to `/d/projects` I get an error message telling me that my username/password is invalid, when I didn't even try to login.

I think if you specify this message param on the loginMiddleware and not the jwtMiddleware that will solve the problem. I think you could have the jwt middleware redirect to `sign-in` without the query param, but I'm not sure if there are scenarios where we won't want that

dheniges (changes requested):

dheniges (approved):
solid

PR #97, opened 2019-04-18
Focus field on creation

jcontonio:
* Focuses and styles field on creation
* Tracks created field in store until an update or destroy happens
* Renames CSS for fields to `document-field` because it clashed with form-fields

### DIFF
--- a/web/src/client/js/reducers/documentFields.js
+++ b/web/src/client/js/reducers/documentFields.js
@@ -3,7 +3,7 @@ import { ActionTypes } from '../actions'
 const initialState = {
   // nested at { [documentId]: { [fieldId]: {} } }
   documentFieldsById: {},
-  lastCreatedField: null,
+  lastCreatedFieldId: null,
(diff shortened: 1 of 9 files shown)


web/src/client/js/reducers/documentFields.js
```diff
@@ -3,6 +3,7 @@ import { ActionTypes } from '../actions'
 const initialState = {
   // nested at { [documentId]: { [fieldId]: {} } }
   documentFieldsById: {},
+  lastCreatedField: null,
```
dheniges: nitpicky feedback, but this would be clearer if named `lastCreatedFieldId`

dheniges (changes requested):
I apologize for the nitpick but I do think it'll make it easier to work with down the road.
I'm not sure how to test the scroll thing, but functionally everything seems to work 👍

dheniges (approved):

PR #98, opened 2019-04-18
add number length validation

jjidt:
- adds a limitation of 15 significant digits for number field values
- sets up a better way to test typed errors using expect.objectContaining


dheniges (approved):
I was assuming there must be a way to check for error types, that's nice 💯

PR #101, opened 2019-04-22
project token coordinator and tests

dheniges:
- Create your very own project token via coordinator!
- Adds a couple more functions to BusinessProjectToken to enable full endpoint support
- Adds random string generator library, used for creating secrets to sign tokens with

### DIFF
--- a/api/src/coordinators/projectToken.test-u.js
+++ b/api/src/coordinators/projectToken.test-u.js
@@ -1,0 +1,44 @@
+import { createProjectToken } from './projectToken'
+import randomStringGenerator from '../libs/randomStringGenerator'
+import tokenGenerator from '../integrators/token'
+import BusinessProjectToken from '../businesstime/projecttoken'
+import resourceTypes from '../constants/resourceTypes'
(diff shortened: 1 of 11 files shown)


api/src/coordinators/projectToken.test-u.js
```diff
@@ -0,0 +1,44 @@
...
+
+  describe('#createProjectToken', () => {
+    let token
+
+    beforeAll(async () => {
+      token = await createProjectToken({ name, projectId, roleId }, 1)
+    })
+
+    it('should call BusinessProjectToken.create', () => {
+      expect(BusinessProjectToken.create.mock.calls.length).toBe(1)
```
jjidt, line 25: should probably test that this and the following few tests are getting called with the correct args
we can get the mock return value from rsg with something like

`randomStringGenerator.mock.results[0].value`

jjidt (changes requested):

api/src/coordinators/projectToken.test-u.js
dheniges: updated tests

jjidt (approved):

PR #104, opened 2019-04-22
field order

jjidt:


api/src/businesstime/field.js
```diff
@@ -92,6 +92,51 @@ async function deleteByIdForDocument(id, docId, orgId) {
...
+    throw ono({ code: 400, message }, message)
+  }
+
+  const field = await db.model(MODEL_NAME).findOne({ where: { id, docId, orgId } })
+  if (!field) throw ono({ code: 404 }, `Cannot update order, field not found with id: ${id}`)
+
+  const currentPosition = field.order
+
+  // no order position change, do nothing
+  if (newPosition === currentPosition) return
```
dheniges, line 109: good call

api/src/businesstime/field.js
```diff
@@ -92,6 +92,51 @@ async function deleteByIdForDocument(id, docId, orgId) {
...
+
+  const currentMax = (await db.model(MODEL_NAME).count({ where: { docId, orgId } })) - 1
+
+  if (newPosition > currentMax) {
+    const message = `Cannot update order, the final field order position for this document is ${currentMax}`
+    throw ono({ code: 409, message }, message)
```
dheniges, line 115: why is message passed twice? Is that something we should be doing with ono in order to capture the message up the line?

api/src/businesstime/field.js
```diff
@@ -92,6 +92,51 @@ async function deleteByIdForDocument(id, docId, orgId) {
...
+  try {
+    transaction = await db.transaction()
+
+    if (newPosition > currentPosition) {
+      // moving to a higher order position
+      await db.query(`UPDATE "Fields" SET "order" = "order" - 1 WHERE "order" >= ${currentPosition} and "order" <= ${newPosition};`, { transaction })
+      // await db.query(`UPDATE "Fields" SET "order" = "order" + 1 WHERE "order" > ${newPosition};`, { transaction })
+    } else if (newPosition < currentPosition) {
+      // moving to a lower order position
+      await db.query(`UPDATE "Fields" SET "order" = "order" + 1 WHERE "order" >= ${newPosition} and "order" < ${currentPosition};`, { transaction })
```
dheniges, line 129: oh these are nice. I did complex calculations by moving my fingers to follow the logic. Finger logic reasoning says it's solid 💯

dheniges (approved):
It looks good and it orders. The only problem I see is adding fields doesn't add order, so we'll need to do that to ensure orders are sequential with no gaps

api/src/businesstime/field.js
jjidt: the message in the object will get passed to the api consumer, the second arg message is what we're logging internally, which here is the same thing. I kind of like the redundancy, just to keep any pattern of directly sending an error message out of the controller.

PR #105, opened 2019-04-23
Project API Token endpoints

dheniges:
Requires #95 

- Adds controller for tokens
- Updates permissions to handle any request with a project id for project resources

Manually tested endpoints in postman


jjidt (approved):
👍

PR #107, opened 2019-04-24
add order on field creation

jjidt:
- normalizes field order on create/delete/order updating
- adds next order position to created fields


dheniges (approved):
I added new fields and they ordered. Huzzah!

PR #108, opened 2019-04-24
UI field ordering to endpoint and reducer

jcontonio:
* Adds field order action / reducer
* Hits endpoint to re-order a field

This works, but I'm mutating state in the reducer and don't necessarily know the best way to not do that 🤷‍♂️


web/src/client/js/reducers/documentFields.js
```diff
@@ -55,40 +55,44 @@ export const documentFields = (state = initialState, action) => {
       return { ...state, documentFieldsById, lastCreatedFieldId, loading: { ...state.loading, byId: loadingById } }
     }
     case ActionTypes.INITIATE_FIELD_ORDER: {
+      // No change, early return
+      if (oldOrder === newOrder) return state
```
jcontonio: I have this in the component, do you think it's worth removing from there so that at least the action triggers?

jjidt: In this case, I think it can be in both places

web/src/client/js/reducers/documentFields.js
```diff
@@ -55,40 +55,44 @@ export const documentFields = (state = initialState, action) => {
...
+
       const { documentId, fieldId, newOrder } = action
       const documentFields = { ...state.documentFieldsById[documentId] }
-      const fieldToUpdate = { ...state.documentFieldsById[documentId][fieldId] }
+      const { [fieldId]: fieldToUpdate, ...remainingFields } = documentFields
```
jcontonio, line 59: Oh this is nice to get the field and remaining fields... very nice

web/src/client/js/reducers/documentFields.js
```diff
@@ -1,4 +1,5 @@
 import { ActionTypes } from '../actions'
+import { filterObject } from 'Shared/utilities'
```
dheniges: Looks like this isn't used?

web/src/shared/utilities.js
```diff
@@ -23,6 +23,15 @@ export const throttle = function(action) {
   }
 }
 
+export const filterObject = (obj, predicate) => {
```
dheniges: Remove this function? Or document it (looks like it takes a function as a second arg that's supposed to return a boolean when given the value of an object's key?)

dheniges (changes requested):
Looks like a leftover util function that can maybe be removed

web/src/client/js/reducers/documentFields.js
jjidt: yeah, I was on the fence on that one, since it's maybe a useful util

YAGNI though, we can add it when it's used

dheniges (approved):

PR #109, opened 2019-04-26
org permission component

jjidt:
- adds the OrgPermission component which accepts an array of acceptedRoleIds and an elseRender block of JSX
- additional higher-order component `withOrgPermission` is also exported, which allows us to wrap entire components at export
- adds `READ_MY` perm for users, so third tier users can now fetch their own user
- currently just set up with Project Creation in the UI, removing the add new project links from the toolbars, and blocking out the create component for users without permission

NOTE: I'm passing in [1, 2] in 3 different places for acceptedRoleIds, do we want some kind of reverse-perms constants file to DRY that up? `CREATE_PROJECT: [1, 2]` ?


dheniges (changes requested):
Yeah I'd add constants for the roles like we have on the API. `PROJECT_ROLE_IDS.READER` etc.

Otherwise looks nice 👍

jjidt:
updated, ready for another look

dheniges (approved):
sweet

PR #111, opened 2019-04-26
project token auth

dheniges:
So you have a project token? And you want to use it to access project resources? Now you can!!!

- custom passport strategy to use project tokens
- code path in permissions is pretty quick to skip over orgs, and project user roles when using project tokens, so little overhead there
- The most overhead is having to run the jwt auth and fail before running the custom project token jwt auth. Might be a way to short-circuit that down the road

### DIFF
--- a/api/src/businesstime/project.test-i.js
+++ b/api/src/businesstime/project.test-i.js
@@ -112,35 +112,6 @@ describe('BusinessProject', () => {
         })
       })
     })
-
-    describe('#findByIdUnsanitized', () => {
-      let targetProjectId
-      let project
-
-      describe('when the target project has the passed in orgId', () => {
-        beforeAll(async () => {
-          targetProjectId = factoryProjects[0].get('id')
-          project = await BusinessProject.findById(targetProjectId, constants.ORG_ID)
-        })
-
-        it('should return a project as POJO', () => {
-          expect(project.id).toBe(targetProjectId)
-          expect(project.constructor).toBe(Object)
-          expect(Object.keys(project)).toEqual(expect.arrayContaining(PUBLIC_FIELDS))
-        })
-      })
-
-      describe('when the target project does not have the passed in orgId', () => {
-        beforeAll(async () => {
-          targetProjectId = factoryProjects[0].get('id')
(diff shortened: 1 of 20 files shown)


api/src/businesstime/project.test-i.js
```diff
@@ -112,6 +112,35 @@ describe('BusinessProject', () => {
...
+      describe('when the target project has the passed in orgId', () => {
+        beforeAll(async () => {
+          targetProjectId = factoryProjects[0].get('id')
+          project = await BusinessProject.findById(targetProjectId, constants.ORG_ID)
+        })
+
+        it('should return a project as POJO', () => {
+          expect(project.id).toBe(targetProjectId)
+          expect(project.constructor).toBe(Object)
+          expect(Object.keys(project)).toEqual(expect.arrayContaining(PUBLIC_FIELDS))
```
jjidt: should we also check for at least one non-public field?

api/src/businesstime/projecttoken.test-i.js
```diff
@@ -119,6 +119,27 @@ describe('BusinessProjectToken', () => {
...
+      factoryToken = (await TokenFactory.createMany(1, {
+        projectId: 123,
+        orgId: constants.ORG_ID
+      }))[0]
+
+      token = await BusinessProjectToken.findByIdUnsanitized(factoryToken.id, factoryToken.orgId)
+    })
+
+    it('should return public fields', () => {
+      expect(Object.keys(token)).toEqual(expect.arrayContaining(PUBLIC_FIELDS))
```
jjidt, line 135: same here, do we want to have a check for a non-public field

jjidt (commented):
manually tested with multiple tiers and it works great! great tests, just one suggestion

api/src/businesstime/projecttoken.test-i.js
dheniges: the next test checks for the secret

api/src/businesstime/project.test-i.js
dheniges: I accidentally copied and didn't remove these tests, put them in the wrong file. Deleted them now.

jjidt (approved):

PR #113, opened 2019-05-01
Global icon and font-size adjustments

jcontonio:
* Adjusts to use the 62.5% base font and rem values everywhere accordingly
* Adjusts icons to default to 18x18, and 12x12 if they are in a button w/ circular icon
* Adds a variable for icon margin
* Adjusts bulma size variables in theme


dheniges (approved):
Clicked around a bunch and nothing seemed off 👍

PR #114, opened 2019-05-01
Fields fetch uses document instead

jcontonio:
* Removes request / receive FIELDS actions and fetching
* Switches documentFields reducer to use document receive action
* Needs updated document endpoint with values to work onload... getting everything on update, etc.


dheniges:
holding off on this until there are field values on this endpoint so it doesn't break the app

dheniges:
@jcontonio the documents endpoint has been updated with values now in master, but there's a conflict here

jcontonio:
@jjidt this is ready for ya

jjidt (approved):
👍 looks good

I fixed the merge conflict
also fixed a bug with the fields list not updating when going from a single field to no fields

PR #115, opened 2019-05-02
publish coordinator

dheniges:
- It versions a document and copies all its fields and values
- Coordinator tests are integration because I was concerned I was just making up stuff and the publish wasn't actually happening 🤷‍♂  I can add unit tests as well if that sounds useful
- DocumentVersion gets Business-y

### DIFF
--- a/api/src/businesstime/documentversion.test-i.js
+++ b/api/src/businesstime/documentversion.test-i.js
@@ -1,0 +1,35 @@
+import BusinessDocumentVersion from './documentversion'
+import ProjectFactory from '../db/__testSetup__/factories/project'
+import DocumentFactory from '../db/__testSetup__/factories/document'
+import initializeTestDb from '../db/__testSetup__/initializeTestDb'
+
+describe('BusinessDocument', () => {
+  let factoryDocument
+
+  beforeAll(async () => {
+    await initializeTestDb()
+    const factoryProjects = await ProjectFactory.createMany(1)
+    const factoryProject = factoryProjects[0]
+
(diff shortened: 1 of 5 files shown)


api/src/businesstime/documentversion.test-i.js
```diff
@@ -0,0 +1,50 @@
...
+
+    it('should belong to the document', () => {
+      expect(documentVersion.docId).toBe(factoryDocument.id)
+    })
+
+    it('should belong to the org', () => {
+      expect(documentVersion.orgId).toBe(factoryDocument.orgId)
+    })
+
+    it('should have a verison number', () => {
```
jjidt: typo

api/src/coordinators/documentVersion.js
```diff
@@ -0,0 +1,25 @@
+import BusinessDocument from '../businesstime/document'
+import BusinessField from '../businesstime/field'
+import BusinessDocumentVersion from '../businesstime/documentversion'
+
+const publishDocumentVersion = async function(docId, projectId, orgId) {
+  const doc = await BusinessDocument.findByIdForProject(docId, projectId, orgId)
```
jjidt, line 7: do we need a check here that the document exists and a 404 if not? might be coming back null

jjidt: probably a test for that situation too

jjidt (changes requested):
I like the integration tests here, and I don't think we need unit tests as long as we have these  💯 

just a couple small changes

dheniges:
@jjidt updated

jjidt (approved):

PR #116, opened 2019-05-03
Project Tokens UI

jcontonio:
* Lists project tokens
* Ability to select a key and see project endpoints examples change token
* Adds new ClipboardComponent for copying the text of a passed ref


dheniges (approved):
This is awesome! I dig the UI, also about damn time native copy is supported.

I had to go re-read the hooks docs to make sure the clipboard component was cool without useEffect, and it is.

PR #117, opened 2019-05-03
project perms component

jjidt:
- adds ProjectPermission component
- hooks up project.accessLevel field
- hooks up project setting in UI for giving project default read access
- endpoint for fetching project assignments for user


web/src/client/js/components/Permission/ProjectPermission.js
```diff
@@ -0,0 +1,57 @@
...
+    </>
+  )
+
+  const { data: project = {} } = useDataService(dataMapper.projects.id(projectId))
+  const { data: userProjectAssignments = {} } = useDataService(dataMapper.users.currentUserProjectAssignments())
+
+  const projectAssignment = userProjectAssignments[projectId]
+
+  // Check the manual project role assignments for the current user
+  if (projectAssignment && acceptedRoleIds.indexOf(projectAssignment.roleId) > -1) {
```
dheniges: `indexOf` is fine but I like the readability of `includes`

dheniges (changes requested):
There's a console warning on the checkbox, it wants a string not a boolean value:
Screen Shot 2019-05-06 at 10 44 27 AM

In terms of names on endpoints, we have `/projects/:id/assignments` and then `/users/:id/projectassignments` though both return an "assignment." In the case of the users/ endpoint it adds a bit of clarity, but on the projects/ endpoint it would be repetitive to call it `projects/:id/projectassignments`. I'm leaning toward wanting to call them both "assignments" but what do you think?

jjidt:
updated, I'm into the /assignments consistency 👍

dheniges (approved):

PR #118, opened 2019-05-03
attach field values on document fetch

jjidt:


dheniges (approved):
👍 sweet that include is reusable for field values

PR #119, opened 2019-05-06
Org owner sign-up flow

jcontonio:
organization-flow@2x


dheniges:
I think I'm following what's going on here. To clarify one piece:
1. User signs up, creates org
1. User has _not_ verified email
1. User has 10 days to verify email before account is locked

If the above is true, we'll have to see how many spam signups we get. If the number isn't high (not sure what "high" means, more than 10/week or something?), this looks good. If we're getting thousands of spam signups per week then we may need to lock it down more and require email verification up front

dheniges:
Thinking about the org owner, can they add users before they've picked a plan/paid? I'm thinking the answer should maybe be "no"?

jcontonio:
> I think I'm following what's going on here. To clarify one piece:
> 
> 1. User signs up, creates org
> 2. User has _not_ verified email
> 3. User has 10 days to verify email before account is locked
> 
> If the above is true, we'll have to see how many spam signups we get. If the number isn't high (not sure what "high" means, more than 10/week or something?), this looks good. If we're getting thousands of spam signups per week then we may need to lock it down more and require email verification up front

The only reason I had that countdown in there was because I thought we may *have* to create an org on the back-end for any of this to work? If it's possible to not create the org until the user is verified, that's the better option.

jcontonio:
> Thinking about the org owner, can they add users before they've picked a plan/paid? I'm thinking the answer should maybe be "no"?

Right, I don't think they should be able to. I can adjust the diagram to make that clearer than the whole Users / Projects admin view is only really available for certain plans

jcontonio:
Unless we give them a limit of 5 users? What if someone wants to evaluate the whole tool?

jjidt:
hmm.. thinking through how the flow would work if we didn't require an organization to be created until the user is verified. I think? that would be fine, allowNull on the orgId and store the signup token on the user. Send it via email, and then require them to create an org on verification.

jcontonio:
I am totally down with that if y'all are

jjidt:
I'm also leaning toward no for managing users on a preview plan.  If people want to evaluate they can share creds, and that might be an early incentive to sign up for a paid plan.

dheniges:
👍 to forcing verification and not creating an org if a user is doing a trial until the user verifies their email

jcontonio:
Ok, so other than making it more clear that Users / Projects admin view is only available to certain plans, this diagram look good to you? Suggest any edits?

dheniges (approved):

PR #120, opened 2019-05-06
Project tokens creation/deletion

jcontonio:
* Adds ability to create/delete tokens
* Fixes endpoint listing when no token exists
* Adjusts DATA docs for clearer steps when needing to re-run migrations


web/src/client/js/actions/index.js
```diff
@@ -88,6 +93,10 @@ export const ProjectAssignees = {
 export const ProjectTokens = {
   request: makeActionCreator(ActionTypes.REQUEST_PROJECT_TOKENS, 'projectId'),
   receive: makeActionCreator(ActionTypes.RECEIVE_PROJECT_TOKENS, 'projectId', 'data'),
+  create: makeActionCreator(ActionTypes.CREATE_PROJECT_TOKEN, 'projectId'),
+  receiveOneCreated: makeActionCreator(ActionTypes.RECEIVE_CREATED_PROJECT_TOKEN, 'projectId', 'data'),
```
jjidt: the projectId is in the returned data. Do we need to pass the projectId in the action as well, or can reducers pull that out of the data object?

jjidt (changes requested):
Looks/works great! just the one nitpick

jcontonio:
All set, thanks @jjidt

jjidt (approved):
👍

PR #122, opened 2019-05-07
Clearing out user assignment on project create

jcontonio:
* Clears out userAssignment for the user on project create


dheniges (approved):
👍

PR #124, opened 2019-05-08
UI – Adding in Permission component places

jcontonio:
* Updates projectId type on ProjectPermission component
* Adding in permissions check to Navigator, ProjectListItem, and ProjectSettings


jjidt:
@jcontonio pushing this back to you -

- fixed loading state issues with user-assignments and projects
- sets an internal mounted flag in useDataService to prevent setting internal state if the component is unmounted

web/src/client/js/libs/makeCancelable.js
```diff
@@ -0,0 +1,22 @@
+const makeCancelable = (promise) => {
+  let hasCanceled_ = false
```
jcontonio: what's this underscore indicate?

jjidt: wait, didn't I delete this file?

jjidt: that was copied code from a blog post

jcontonio: Oh, yeah you did in the following commits.

jjidt (approved):

PR #125, opened 2019-05-08
Publish endpoint + queries

dheniges:
- Adds ability to add `?draft=true` flag to `/documents/:id` and `/documents/:id/fields` to get draft fields. Otherwise endpoints return most recent published fields.
- Adds publish endpoint

- Field ordering needs to be updated to handle published/non-published fields
- UI updates needed


api/src/businesstime/field.test-i.js
```diff
@@ -181,25 +185,50 @@ describe('BusinessField', () => {
...
       })
-      await DocumentFactory.createMany(2, {
-        docId: 0,
-        orgId: constants.ORG_ID
+
+      it('should return fields as POJOs', () => {
+        for (const field of fields) {
+          expect(field.constructor).toBe(Object)
+          expect(Object.keys(field)).toEqual(expect.arrayContaining(resourcePublicFields[resourceTypes.FIELD]))
+          expect(Boolean(field.value)).toBe(true)
```
jjidt: might be more explicit with `expect(field.value).toBeDefined())` instead?

api/src/db/__testSetup__/factories/document.js
```diff
@@ -15,7 +15,10 @@ const createMany = async function(numberOfDocuments, override) {
   const db = getDb()
   const projects = Array(numberOfDocuments).fill().map(() => getDocumentData(override))
 
-  return Promise.all(projects.map(projectData => db.model('Document').create(projectData)))
+  return Promise.all(projects.map((projectData) => {
+    // console.info(projectData)
```
jjidt: leftover ?

jjidt (changes requested):
Tests look great! really cool to see all of this coming together

Seems like field updating might not be working yet. If I publish a version, go back and change a field value, and publish again the field isn't updated in the published field return

jjidt:
or is that field update covered in this ticket?

`Update /fields/:id endpoint to prevent deleting or editing a published field`

dheniges:
@jjidt ready for another peek. Re: update, if you update or add a field, the next publish should correctly grab that field/update. I manually tested both those scenarios and they worked as expected. I looked into testing it but it's sort of more an endpoint-level integration test.

jjidt (approved):
ah, I was previously testing with the UI which must not be updated to get the draft fields, so it was probably updating the published fields. Tested it with Postman and it looks good 👍

dheniges:
oh yeah, did I mention this is all API side so far? :D

PR #129, opened 2019-05-08
Notifications & Errors

jcontonio:
* Adds new Notifications store/actions/components
* Adds error handling for `fetch` on projects and documents
* Adds in ErrorBoundary to catch any React errors so the user doesn't get a blank screen
  * Added a commented function in there for when we ever want to log app errors somewhere
* Getting eyes on this before going all the way with all the things (post, put, remove, etc)

To test:

* Log in as bobx@bong.com, and you'll see he has no permissions to [here](http://localhost:3000/d/project/110000) and [here](http://localhost:3000/d/projects)
* Break a component by not returning anything (remove a `null` from a `return null` in the UserMenu), and refresh the page

Questions:
* It'd be cool to have a unique message per resource, notice the following when a user doesn't have the project or documents permissions. How can we provide messages like "You don;'t have access to these documents" and "Sorry, this project is not available to you" sort of things?

Screen Shot 2019-05-08 at 2 54 13 PM

### DIFF
--- a/web/src/client/js/api/index.js
+++ b/web/src/client/js/api/index.js
@@ -7,7 +7,7 @@ const token = getCookieValue('token')
 // process.env.API_PUBLIC_URL
 // eslint-disable-next-line no-undef
 const baseURL = new URL('api/', VAR_API_URL)
-const globalErrorCodes = [403, 404, 409, 500]
+const globalErrorCodes = [403, 404, 500]
 
 const axiosInstance = axios.create({
   baseURL: baseURL.toString(),
@@ -19,10 +19,11 @@ const axiosInstance = axios.create({
 
 async function fetch(resource) {
   try {
-    const { data } = await axiosInstance.get(resource)
-    return data
+    const { data: { data }, status } = await axiosInstance.get(resource)
+    return { data, status }
   } catch (error) {
-    return error.response
+    const { data, status } = error.response
+    return { data, status }
   }
 }
 
--- a/web/src/client/js/api/index.js
+++ b/web/src/client/js/api/index.js
@@ -7,6 +7,7 @@ const token = getCookieValue('token')
 // process.env.API_PUBLIC_URL
 // eslint-disable-next-line no-undef
 const baseURL = new URL('api/', VAR_API_URL)
+const globalErrorCodes = [403, 404, 409, 500]
 
 const axiosInstance = axios.create({
   baseURL: baseURL.toString(),
@@ -17,8 +18,12 @@ const axiosInstance = axios.create({
 })
 
 async function fetch(resource) {
-  const { data } = await axiosInstance.get(resource)
-  return data
+  try {
+    const { data } = await axiosInstance.get(resource)
+    return data
+  } catch (error) {
+    return error.response
+  }
 }
 
 async function add(resource, body) {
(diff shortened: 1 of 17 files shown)


web/src/client/js/api/index.js
```diff
@@ -7,6 +7,7 @@ const token = getCookieValue('token')
 // process.env.API_PUBLIC_URL
 // eslint-disable-next-line no-undef
 const baseURL = new URL('api/', VAR_API_URL)
+const globalErrorCodes = [403, 404, 409, 500]
```
jjidt: Not sure if we want 409s as globals or not, I think it's fine for now but we might want to revisit

web/src/client/js/components/Notifications/NotificationComponent.js
```diff
@@ -0,0 +1,61 @@
...
+    'notifications__notice': true,
+    'notifications__notice--error': notice.type === 'error',
+    'notifications__notice--success': notice.type === 'success',
+    'notifications__notice--warning': notice.type === 'warning'
+  })
+  const noticeRef = useRef()
+  const title = `${id}-title`
+  const desc = `${id}-desc`
+  let noticeTitle
+  switch (notice.code) {
```
jjidt: are error notices the only ones that will have a code? If so I'd say put this in a conditional and break out the switch into a helper function outside of the NotificationComponent function

```
if (notice.type === NOTICE_TYPES.ERROR) {
  noticeTitle = getErrorTitle(notice.code)
}
```

web/src/client/js/components/Notifications/NotificationComponent.js
```diff
@@ -0,0 +1,61 @@
+import React, { useRef } from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+
+import { RemoveIcon } from 'Components/Icons'
+
+const NotificationComponent = ({ dismissHandler, id, notice }) => {
+  const noticeClasses = cx({
+    'notifications__notice': true,
+    'notifications__notice--error': notice.type === 'error',
```
jjidt: let's defined each notice.type in a constants file

web/src/client/js/reducers/notifications.js
```diff
@@ -0,0 +1,38 @@
+import { uniqueId } from 'lodash'
+import { ActionTypes } from '../actions'
+
+const initialState = {
+  notices: {}
```
jjidt: any reason for using both notice and notification? could we use one or the other?

web/src/client/js/api/index.js
```diff
@@ -17,8 +18,12 @@ const axiosInstance = axios.create({
 })
 
 async function fetch(resource) {
```
jjidt, line 20: lets make the return object consistent, whether its a 200 level code or an error

```
async function fetch(resource) {
  try {
    const { data: { data }, status } = await axiosInstance.get(resource)
    return { data, status }
  } catch (error) {
    const { data, status } = error.response
    return { data, status }
  }
}
```

jjidt (changes requested):
go to
`http://localhost:3000/d/project/110002` and you'll get a console error about passing an object as the message field to notifications, so I need to go through and make those consistent, 403s have the message directly on the `data` field, others have `data.error` for the message

PR #130, opened 2019-05-09
error consistency

jjidt:
- gets 403s in line with other error types
- started to add a ProjectPermissionError type, but realized all permissions are passthrough Organization check -> Project check so there's not a good way to differentiate these since the response will always be sent in the Project check
- updated docs with full error format


jcontonio (dismissed):
Looks good here

PR #132, opened 2019-05-10
express next() error handling

jjidt:
- handle api errors with express middleware
- 404 all perm errors
- pass generic public messages


jcontonio (approved):
Looks good to me clicking around 👍

dheniges (approved):
Looks good :+1: that error handling is much nicer, and I like keeping the try/catch on the controller to catch whatever comes up below still rather than having to rely on correctly calling an error handler callback everywhere else.

PR #135, opened 2019-05-14
CRUD users

jjidt:


dheniges (commented):
Code looks good. I'm going to wait for you to merge in the name update here and remove that TODO and test it out

jjidt:
@dheniges updated

dheniges (commented):
Functionality works as expected 👍 

For POST users though, there's no return. I would expect it to return the user that's created

jjidt:
updated with user return

jjidt:
ooh, should probably add that to the coordinator test

dheniges (approved):
Huzzah, users!

PR #137, opened 2019-05-15
soft delete

jjidt:
- updates id references in migration files
- moves org migration before user migration
- adds paranoid: true to all models which results in deletedAt timestamp being added to soft-deleted items


jjidt:
oops, I broke all the tests ha

jjidt:
locking in the id references exposed a bunch of false positives in the tests, all fixed up now

dheniges (commented):
I tested deleting a lot of things and all seemed well 👍 

I looked around a bit and beyond ordering (which I'm working on, and will fix) I didn't see any custom querying. Any custom queries you can think of that will need to respect deleted resources?

jjidt:
nope, I think ordering is all

dheniges (approved):
Stamp of Approval™

PR #138, opened 2019-05-16
UI – Page titles

jcontonio:
* Adds page titles
* Adding in react-helmet package to manage <meta> tags throughout the app
* Added <Helmet>'s to containers specifically, to keep them separate & obvious

`make installweb`


dheniges:
@jcontonio conflict funtimes!

jcontonio:
@dheniges fixed up

dheniges:
The way this is setup it will keep the last page title set. Is there a way to set a default title with helmet that will get set if nothing else in the current render is updating the title?

For example, if I go to a document I see "Project Name: Document Name" which is cool. But then if I navigate to the user menu and go to settings, there's no page title set so it keeps the "Project Name: Document Name" title.

dheniges (approved):
We'll see what codefresh has to say, but looks good to me

PR #139, opened 2019-05-16
Publish order

dheniges:
- Merges in soft delete updates from master
- Fixes ordering to only order draft fields
- Updates client to load draft fields


jjidt (approved):
Works great! put it through some manual testing and it seems to be working how we want 👍

PR #140, opened 2019-05-16
fix project perm bug

jjidt:
- for projects, if we try to fetch something that does not exist, or ends up erroring out, we still need to update the loading status in the reducer
- we also need to retain the loading status for those non-existent items even when the projects list comes back
- this PR adds those things and also includes an additional `mounted` check in useDataService to prevent a race condition where the component unmounted before the loading state could change


dheniges:
So if I don't have access to a project, this is expected?
Screen Shot 2019-05-17 at 10 19 38 AM

jjidt:
yeah for now
we don't have a redirect on that yet, and I was asking jay about whether we want those error messages yesterday
but that's not related to the bug

dheniges (approved):
It works!

PR #141, opened 2019-05-17
UI Publish + Lockdown Published Fields

dheniges:
- Prevents deleting published fields
- Prevents updating published fields
- Plugs in UI publish button to publish endpoint
- Fixes an ordering bug with publish

### DIFF
--- a/api/src/businesstime/field.test-i.js
+++ b/api/src/businesstime/field.test-i.js
@@ -167,7 +167,7 @@ describe('BusinessField', () => {
       it('should throw an error', async () => {
         await expect(
           BusinessField.updateByIdForDocument(publishedField.id, docId, orgId, updateBody)
-        ).rejects.toThrow()
+        ).rejects.toEqual(expect.objectContaining({ code: 403 }))
       })
     })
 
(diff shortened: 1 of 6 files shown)


dheniges:
@jjidt ready for review whenever you get time today

api/src/businesstime/field.test-i.js
```diff
@@ -150,6 +150,27 @@ describe('BusinessField', () => {
...
+          docId: updateFactoryDocument.id,
+          type: factoryField.type,
+          versionId: version.id
+        }))[0]
+      })
+
+      it('should throw an error', async () => {
+        await expect(
+          BusinessField.updateByIdForDocument(publishedField.id, docId, orgId, updateBody)
+        ).rejects.toThrow()
```
jjidt: we should probably check the error code here

.rejects.toEqual(expect.objectContaining({ code: 403 }))

api/src/businesstime/field.test-i.js
```diff
@@ -331,6 +352,28 @@ describe('BusinessField', () => {
...
+          docId: documentToDelete.id,
+          type: fieldC.type,
+          versionId: version.id
+        }))[0]
+      })
+
+      it('should throw an error', async () => {
+        await expect(
+          BusinessField.deleteByIdForDocument(publishedField.id, documentToDelete.id, constants.ORG_ID)
+        ).rejects.toThrow()
```
jjidt: same

jjidt (changes requested):
Fun to see this working, everything looking good, just a couple specific test value checks

jjidt (approved):
nice 👍

PR #143, opened 2019-05-17
handle bodyParser errors

jjidt:


dheniges (approved):
:+1: nice code comments, would not want to trace that out without comments

PR #147, opened 2019-05-22
UI – Owner sign-up flow

jcontonio:
* Adds org name to "completion" page when coming back from an email
* Adds Admin section, and UPGRADE notice in the header
* Needs a `PlanPermission` component

http://localhost:3000/sign-up/processing
http://localhost:3000/sign-up/registration/complete


jjidt (changes requested):
Getting a 404 on the password page when going through the signup process

jjidt (dismissed):
👍 added a thing to retain the orgName on error, and the projectCreation setting when that gets in there

give it a look before merging

jjidt (approved):
Looks good, we do have two different password mismatch error messages though

so maybe make the server fallback one match the UI validation one 'Your passwords don't match' before merging

PR #149, opened 2019-05-22
fix perm race condition

jjidt:
in useDataService, we needed to check the loading and conditionally set the data state before setting the loading state in the component

this was causing problems with having an updated loading state, but non-updated data


jcontonio (approved):
Works 👍

PR #150, opened 2019-05-23
UI – Account Settings View

jcontonio:
* Adds Account section at /settings
  * Makes so only OWNER can see Billing and General Settings
  * Adds in Seats component to manage seats on the org
  * Updates Billing component to manage/display the card on file

* Adds Administration section at /admin
  * Makes so only OWNER and ADMIN can see Users (not in here)

* Adds in new Panel, PanelNavigation, and PanelContent components for consistent UI for things like this (Project Settings)

## Needs

* Org endpoint for name, and project setting
* Billing endpoint hookup (get card display info, update payment info)


web/src/client/js/components/User/UserOrganization/UserOrganizationComponent.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import TextField from 'Components/Form/TextField'
+import Checkbox from 'Components/Form/Checkbox'
+
+const UserOrganizationComponent = ({ errors }) => {
+  const helpText = "Checking this box allows anyone in your organization to view this project's documents, whether they are part of the project team or not"
```
jjidt: I don't think this message is correct, should be something about creating projects

jjidt (changes requested):
Looks good, just the one text change

jjidt (approved):
👍

PR #152, opened 2019-05-24
org update

jjidt:
add organization update route, pretty straightforward

bonkey@bong.com should have access to org 100000, but no one else


jcontonio (approved):
Can update orgs 👍

PR #153, opened 2019-05-24
UI – Users listing/creation

jcontonio:
* Adds new Table component that is flexible grid instead of <table> (for future responsiveness)
* Adds user listing
* Adds user creation
* Adds user deletion

@jjidt can you see why the list isn't being updated after the the ADMIN update action? Also, User's emails are being set to their names, and I can't catch what's going on


jcontonio:
Looks like the name/email thing is API side

image

jjidt (changes requested):
getting a couple console errors in the client:

`Failed prop type: The prop `action.callback` is marked as required in `ToolbarComponent`, but its value is `undefined``

`Failed prop type: The prop `action.title` is marked as required in `ToolbarComponent`, but its value is `undefined``

web/src/client/js/components/Toolbar/ToolbarComponent.js
```diff
@@ -21,7 +21,7 @@ const ToolbarComponent = ({ action, filter, sort }) => {
           className="btn btn--blank btn--with-circular-icon"
           onClick={action.callback}
           title={action.title}>
-          {action.icon}
+          {action.icon && (action.icon)}
```
jjidt: duplicate ?

web/src/client/js/actions/user.js
```diff
@@ -29,6 +60,16 @@ export const updateUser = (userId, body) => {
   }
 }
 
+export const updateUserRole = (userId, body) => {
+  return async (dispatch) => {
+    dispatch(Users.initiateUpdate(userId))
+    const { data, status } = await update(`users/${userId}/orgRole`, body)
+    validationCodes.includes(status) ?
+      dispatch(Validation.create('user', data, status)) :
+      dispatch(Users.receiveOneUpdated(data))
```
jjidt: This won't work how we want. The orgRole endpoint will return a 204 on success with no data, so we'll need a specific action for updating the user role, and we'll need to set the new role on the existing user in the User reducer.

jjidt (changes requested):

web/src/client/js/actions/index.js
```diff
@@ -65,6 +65,11 @@ export const ActionTypes = {
   RECEIVE_USER: 'RECEIVE_USER',
   INITIATE_USER_UPDATE: 'INITIATE_USER_UPDATE',
   RECEIVE_UPDATED_USER: 'RECEIVE_UPDATED_USER',
+  RECEIVE_UPDATED_ROLE: 'RECEIVE_UPDATED_ROLE',
```
jjidt: sorry, nitpicky, but can you make this RECEIVE_UPDATED_USER_ROLE to avoid the possibility of a namespace issue down the road

web/src/client/js/reducers/users.js
```diff
@@ -27,22 +27,48 @@ export const users = (state = initialState, action) => {
...
-    case ActionTypes.RECEIVE_USER: {
-      const id = action.data.id
-      const usersById = { ...state.usersById, [id]: action.data }
-      const loadingById = { ...state.loading.byId, [id]: false }
-      return { ...state, usersById, loading: { ...state.loading, byId: loadingById } }
+    case ActionTypes.RECEIVE_UPDATED_ROLE: {
+      const { userId, orgRoleId } = action
+      const userObj = { ...state.usersById[userId], orgRoleId }
+      const byId = { ...state.loading.byId, [userId]: false }
+      //userObj.orgRoleId = orgRoleId
```
jjidt: leftover

jjidt (changes requested):
Just a couple things,

Also, we probably don't want to allow users to delete themselves in the admin list, I tried it and everything (understandably) broke

jjidt (approved):
🌮

PR #154, opened 2019-05-28
UI – Adds loading state adjustment for project creation

jcontonio:
* Adjusts list loading state when creating project

**Note**: Had to add a check for `projectAssignment` in ProjectPermission component or it would error out with projectAssignment being undefined. Not sure if this is why you can create a project, go to that project, but the ProjectPermission component is still hiding the Settings link from you until you do a hard refresh.


jjidt (approved):
👍 This looks good, I guess it makes sense for that loading state to change on creation, we might need to do that elsewhere

The existing projectAssignment check used to be in there and I think got accidentally removed, nice catch

PR #157, opened 2019-05-29
require name on user create

jjidt:
- also removes org name set on the registration form


jcontonio (dismissed):
👍 Created a new user/org - but also I pushed up a tiny change on the password form

jcontonio (approved):
🍕

PR #158, opened 2019-05-29
UI – Admin User View

jcontonio:
* Adds user view, with route, so an admin can look up a user and see their info
* Mostly for the future when we have logs

This is branched from the #153


jcontonio:
This will need #153 merged in

jjidt (approved):
🥇

PR #159, opened 2019-05-29
user verification

jjidt:
- adds email verification when creating a user through through the users/ POST route
- returns token on initial password set route
- automatically logs in user using this token


jcontonio (approved):
👍 Works nicely

PR #160, opened 2019-05-29
user delete perms

jjidt:
-  also adds a validation enum for org role id updates on the user

### DIFF
--- a/api/src/constants/acceptableRoleIdUpdateValues.js
+++ b/api/src/constants/acceptableRoleIdUpdateValues.js
@@ -1,1 +1,3 @@
-export default Object.freeze([2, 3])
+import { ORGANIZATION_ROLE_IDS } from './roles'
+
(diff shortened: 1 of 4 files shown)


api/src/constants/acceptableRoleIdUpdateValues.js
```diff
@@ -0,0 +1 @@
+export default Object.freeze([2, 3])
```
dheniges: should use the organization role id constants

dheniges (dismissed):
Looks good, just one constant update.

Tested all the deletes and everything worked as expected

api/src/constants/acceptableRoleIdUpdateValues.js
jjidt: ah good call

dheniges (approved):
Re-approved

PR #161, opened 2019-05-30
UI – Project Users

jcontonio:
* Adds project users to project list
* Redesign of project list
* Fixes major project assignments loading bug


web/src/client/js/components/ProjectUsers/ProjectUsersCollapsedItem.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+  return (
+    <li className="project-users__collapsed-item" style={itemStyle} title={user.name}>
+      {firstInitial}
+    </li>
+  )
+}
+
+ProjectUsersCollapsedItem.propTypes = {
+  color: PropTypes.string,
```
jjidt, line 17: since this isn't required, is the default color being set somewhere else?

jcontonio: It's being set in the CSS, though I'll add a default color here

jjidt (approved):
👍

PR #164, opened 2019-05-31
Cast string number field value to a number for API

dheniges:
- Title pretty much says it all
- Works with decimal points, does not respect significant digits


jjidt (approved):
👍

PR #165, opened 2019-06-03
UI – Admin user page w/ role and projects

jcontonio:
* Adds role switching to user detail page
* Adds project list to user detail page


jcontonio:
@jjidt passing to you for the permission updates

jjidt:
@jcontonio passing this back, user projects should be listing correctly in the admin section now

jcontonio:
@jjidt added in everything, however I am getting a weird error in the projectAssignments reducer. Can you take a look? After that, this will be ready for review

* If you un-assign a project from a user in Project Settings, it works fine
* Doing the same action in the Admin User Page, we get an error in the `REMOVE_PROJECT_ASSIGNEE` reducer saying state.assignmentByProjectId is undefined on this line `const { [action.userId]: ___, ...restProjectAssignees } = state.assignmentsByProjectId[action.projectId]`

jjidt:
@jcontonio updated - we needed to handle the assignments in the userAssignments reducer as well, check it out and make sure everything is working ok

dheniges (commented):
Everything looks good except the error handling seems off. If the API returns a 500 on the project team management page when adding a user, an empty item gets added to the list. Seems like this should be caught before dispatching the receive assignment action. Although I don't think that code is in this PR, but it touches on it. So maybe that's a new ticket?

jjidt:
went ahead and updated the error handling

jcontonio:
👍 Fixes the problem - didn't realize the error handling wasn't in place at all - I think you can approve this @jjidt and we'll merge

jjidt (approved):

PR #167, opened 2019-06-04
UI – Project / Document toolbar

jcontonio:
* Requires #165 so we can returnNull() from dataMapper.documents.id() 
* Adds new ProjectToolbar with links to New Project, Project Settings, a list of teammates, and document stats (if there's a document)
* Removes items from Header, and renames Header for consistency


jcontonio:
image

web/src/client/js/components/Icons/SearchIcon.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+function SearchIcon({ className, fill, height, width }) {
+  // xmlns="http://www.w3.org/2000/svg"
```
dheniges: remove?

web/src/client/js/components/ProjectToolbar/ProjectToolbarComponent.js
```diff
@@ -0,0 +1,66 @@
...
+      </ProjectPermission>
+      {projectUsers && projectUsers.length > 0 &&
+      <div className="project-toolbar__team">
+        <span className="project-toolbar__team-label">Your Team:</span>
+        <ProjectUsersContainer projectId={projectId} collapsed />
+      </div>
+      }
+      {document &&
+      <div className="project-toolbar__document-info">
+        Last update: {moment(document.updatedAt).format('ddd, hA')}
```
dheniges: I think this date format needs more info, if it was updated in the last week this works, but if it's older you can't differentiate. Which "Mon, 9AM"?  I lean toward "last updated X hours ago" or "X days ago", potentially with the full date (June 7th, 2019) available on hover. This could go in a new ticket.

dheniges (commented):
I like the bottom toolbar, I wasn't sure about it at first but it sits nicely out of the way while revealing useful info.

I like it having project/doc info, that makes sense. I'm not sure about "New Project" there though, that seems like it fits better in or around the project selector menu at the top. Thoughts?

web/src/client/js/components/ProjectToolbar/ProjectToolbarComponent.js
jcontonio: N,o we can do the date stuff in this ticket. I'm into that suggestion!

jcontonio:
Yeah, I feel it's pretty standard to have some sort of status bar, and I couldn't find a decent place to stick the Project Settings or Team info, without cluttering everything up. 

I'm with you that New Project was nice up by the navigator, however when you start to see it bouncing around with different length project names, it gets a little weird. I could go either way. Was also thinking we could duplicate New Project in the navigator menu itself.

dheniges:
Oh I like putting "New Project" in the navigator menu too

jcontonio:
Fixed up, and adjusted Navigator to look like this:

image

dheniges (approved):
looks good

PR #170, opened 2019-06-06
UI – Resend invite link

jcontonio:
* Adds in Resend link with a spinner 
* Confirms this sends the email because I sent @c86brown 4 emails accidentally


jjidt (approved):
It works! 👍

PR #171, opened 2019-06-07
Bug – Fixes project assignment bug on project creation

jcontonio:
Before: After creating a project, you wouldn't see project admin UI until a hard refresh
After: You see project admin UI


dheniges (approved):
👍  This removes all user assignments, which made me pause at first but I think that's good because if you're starting a new project you may not be jumping back into another project so it's sort of clearing the cache. And when we get user assignment plugged into project creation we'll want it too. So yeah, 💯

PR #172, opened 2019-06-07
Bug – Changing to fetch user

jcontonio:
Switched to fetching the user, instead of looking at the loaded users


dheniges (changes requested):
Refreshing on this page causes an error in the app

jcontonio:
@dheniges not seeing that in any browser - any ideas?

dheniges:
I see it in all browsers. It looks like it's because `userFromRoute` is not defined initially on refresh, and the other useDataService calls are trying to call `userFromRoute.id` on undefined

dheniges (approved):
:+1:

PR #173, opened 2019-06-07
UI – Adding pagination

jcontonio:
* Adds Paginator
* Adds some queryParam helper utilities
* Test on Admin users


jcontonio:
^ these conflicts should be fixed in TableSort PR, merge that first

web/src/client/js/utilities/queryParams.js
```diff
@@ -0,0 +1,20 @@
+export const parseParams = (queryParams) => {
```
jjidt, line 1: this works, but I also wouldn't be opposed to using something like the `query-string` package, so that it's fully tested

thoughts?

jjidt (commented):
This is looking good, I changed the base branch to `feature-pagination`, I think we'll want to try it out with whatever we decide server-side and have another round of testing on that

web/src/client/js/utilities/queryParams.js
jcontonio: Oh duh, I didn't even think of looking for a utility that existed. Yeah, I think we should do that.

jjidt:
@jcontonio gonna go ahead and approve/merge this so I can work off of it in my api branch, the query Params stuff works, so if we need to swap that out later we can

jjidt (approved):

PR #175, opened 2019-06-28
early error if no user deserialized from auth token

jjidt:
The server was hanging if there was no auth token in the header. This makes sure the auth token contains a valid user before assembling requestor info.

@dheniges Not 100% sure this is where the check should happen, but seemed like we'd need to repeat the logic if we tried to do it in the user / api token auth modules


dheniges (approved):
I think this spot is fine. We've attempted to parse the request info and it's not available, therefore it's not a valid request.

PR #176, opened 2019-07-08
user paging

jjidt:
Some of the project-user code on the client is commented out, so we'll need to tackle that next.

### DIFF
--- a/web/src/client/js/components/Admin/AdminUsers/AdminUsersContainer.js
+++ b/web/src/client/js/components/Admin/AdminUsers/AdminUsersContainer.js
@@ -30,7 +30,7 @@ const AdminUsersContainer = ({
 
   // Update the params on state change
   useEffect(() => {
-    history.push({ search: `?sortBy=${sortBy}&sortMethod=${sortMethod}` })
+    history.push({ search: `?sortBy=${sortBy}&sortMethod=${sortMethod}&page=${page}` })
     // @todo handle the sort action here, but not on the first load? that should
     // already be the default
   }, [history, sortBy, sortMethod])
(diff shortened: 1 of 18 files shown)


jcontonio:
Paging works!

Need to update the AdminUsersContainer to set sortBy, sortMethod, and page to the initial query params, and stay that way.

If I try to go to http://localhost:3000/d/admin/users?sortBy=createdAt&sortMethod=DESC&page=4, or if I refresh the page I am on, the page query param goes away, sort resets, but page 4 will be selected. 

Should that be done in a different PR?

jcontonio:
I should clarify, page 1 results are there, but the page controller has 4 selected, even though `page=4` query param was removed

jjidt:
@jcontonio updated with the paginator container now relying only on url state rather than internal state, click through that and see if everything is working as it should

jjidt:
ah, page param still getting removed when going directly to url, looking at that now

jjidt:
ok, updated

web/src/client/js/components/Admin/AdminUsers/AdminUsersContainer.js
```diff
@@ -23,13 +23,14 @@ const AdminUsersContainer = ({
...
   const [sortBy, setSortBy] = useState('createdAt')
   const [sortMethod, setSortMethod] = useState(QUERY_PARAMS.ASCENDING)
 
   // Update the params on state change
   useEffect(() => {
-    history.push({ search: `?sortBy=${sortBy}&sortMethod=${sortMethod}` })
+    history.push({ search: `?sortBy=${sortBy}&sortMethod=${sortMethod}&page=${page}` })
     // @todo handle the sort action here, but not on the first load? that should
     // already be the default
   }, [history, sortBy, sortMethod])
```
jcontonio: missing page as dependency?

jcontonio (approved):
Looks good! Question on the hook dependency. 
Refreshing will need to adjust sorting as well, but we should wait until that feature gets worked on

web/src/client/js/components/Admin/AdminUsers/AdminUsersContainer.js
jjidt: ah probably

PR #178, opened 2019-07-12
user restore

jjidt:
can test by adding a user with your email in Admin, verify the email, delete user in Admin and then re-add


jcontonio:
When adding a user, the name and email don't show up in the list here until a hard refresh:

image

jcontonio (approved):
👍 other than weird display issue that first time, this worked for me

PR #179, opened 2019-07-12
Cleaning up org redux flow

jcontonio:
* Adds in billing fetch / update redux actions
* Adds in UI state for open/close of stripe form
* Adds icons for known Stripe card types

### To test:

1. Use different card numbers from https://stripe.com/docs/testing#cards
2. See different icons for each card
3. Test that there is no card information at all
4. Form should disable and not be able to be submitted twice


jjidt:
code looks good! with a clean migration/seed, I'm not seeing the payment info section show up in Admin, is something not getting set right?

jjidt (changes requested):
Works great for successful changes, however, if you use a card that throws an error, and then click cancel, when you go back to the form there is no box for credit card number

jjidt (approved):
👍

PR #181, opened 2019-07-18
user name search

jjidt:
test by going to a project's settings and typing in the team member add box, will need to probably rate limit it in some way to prevent api spamming, but it is currently caching searches and not re-requesting, so that does help

### DIFF
--- a/api/src/businesstime/user.js
+++ b/api/src/businesstime/user.js
@@ -51,6 +51,20 @@ async function findAllSanitized(orgId, options) {
     ...paginationOptions
   }
 
+  if (options.nameSearch) {
+    query.where = {
+      [db.Op.and]: [
+        {
+          ...query.where
+        }, {
+          name: {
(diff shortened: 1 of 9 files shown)


jjidt:
@dheniges I _think_ sequelize is protecting us from SQL injection on this stuff, but it wouldn't be bad to have a second opinion

jcontonio:
Search looks good, there's a bug with adding a user, the name isn't visible until you do a hard refresh. Is something weird with redux state?

1. Search for a user
2. Add the user
3. See blank name
4. Refresh

image

jjidt:
gonna take care of the project-user bug in a separate PR, it's unrelated to anything with user search, and will require a change to the API

dheniges (approved):
It looks like sequelize escapes the iLike op. The syntax looks odd in the logs, but you can't escape your way out of the `'` enclosed string

👍

PR #182, opened 2019-07-22
include user in project assignment payload

jjidt:


jcontonio:
Looks like we have a couple bugs here:

1. Added a user that already existed, but didn't get a notification or anything... but see the duplicate problem on the server side
2. Can't delete users anymore

jjidt:
Those things weren't related to the project user work, went ahead and fixed em in this PR though. Ready for another look

jjidt:
We were ending up with empty users in the project settings Team list after assigning a new user, add someone here and make sure their name etc. shows up right away

For the user bugs - delete a user in the admin panel, try to add a user without bonkeybong.com email and make sure you get a 503 alert

jcontonio (approved):
Works great 👍

PR #183, opened 2019-07-23
Reorg of Admin section

jcontonio:
Moves Billing, Org settings under the Admin section


jjidt (changes requested):
This looks good and makes sense

Can you update the `Upgrade your account` link to point to the correct settings page?

jjidt (approved):
looks good 👍

PR #184, opened 2019-07-24
hard-delete project users

jjidt:


jcontonio (approved):
👍 can add / remove / add users to a project

PR #186, opened 2019-07-26
add lastPublishedAt to documents

jjidt:


jcontonio (changes requested):
We'll need to add lastPublishedAt in the return for document, documents, and publish endpoint returns. I've added the display in the client, but I'm not seeing the field on any network response after doing a migration.

jcontonio (dismissed):
lol nevermind - I was on the wrong branch

jcontonio (approved):
Added published at UI

jjidt:
CI failure on build stage, not tests, gonna merge

PR #187, opened 2019-07-29
UI – Editor v1

jcontonio:
* Breaks out a document to have a textarea (body)
* Added new design for fields (data)
* Added full-screen editing option

I am going to keep working on the styling of fields here, but figured I'd get this in front of you guys to talk technical stuff, and if you like this pattern. 

1. Every new document has a body
2. Users can add fields and re-order them


jcontonio:
Should be able to create docs that look good:

image

dheniges (approved):
This looks real nice :+1:

Surprisingly, no code comments :D

PR #188, opened 2019-07-31
audit log

jjidt:
- adds an audit logger and a handful of log points

- currently logging json to the console, but we could configure this however with a logging service


dheniges (changes requested):
lookin good JJ. Only change I'd request is making the model strings into constants. ie `DOCUMENT_MODEL: 'Document'` etc.

dheniges (approved):
:+1: nailed those constants

PR #189, opened 2019-08-07
handle stripe card errors

jjidt:
This grabs stripe payment errors and transfers them to the client, if we want our own custom error messages for this, we'll need to go through the error list here and cover whichever ones we think are applicable

https://stripe.com/docs/error-codes


dheniges (approved):
Yay payment errors!

PR #190, opened 2019-08-19
Renaming everything

jcontonio:


Makefile
```diff
@@ -14,7 +14,7 @@ installapi:
 	docker-compose exec api npm install
 
 cleardev:
-	docker-compose down && docker volume rm project-danger_postgres-data
+	docker-compose down && docker volume rm portway_postgres-data
```
dheniges, line 17: hey look we're all going to have the same db volume now because there's no dash in the directory name!

dheniges (approved):
Everything I tested works with this. Only weirdness for us with a directory still named `project-danger` is docker picks up the directory name to use for a prefix for the network, db volume etc. But renaming the clone directory fixes that

PR #191, opened 2019-08-21
image field

jjidt:
Adds the image field type

To test:

get the Portway api dev .env file from 1pass

using Postman - post to http://localhost:3001/api/documents/120001/fields with form data:

Screen Shot 2019-08-21 at 10 16 39 AM

Note: we'll want to namespace by organization(or create organization buckets) /project/document, this is currently just dumping into a single bucket


dheniges:
@jjidt can you fix the conflicts and merge? Running into project-danger/portway issues trying to test

jjidt:
@dheniges updated, just merged in Jay's changes, what I had on the client wasn't actually hooked up yet so didn't really matter

dheniges (commented):
It's functional, and uploads and returns the saved url!

1) I noticed the error handling is weird if the file type parameter in the body is not part of a multpart body/octet-stream. Probably want to add a ticket to make that more robust.

2) Multer's default memory store is problematic for large files / many files uploaded at once. Best practice is to write it to disk and delete it after it gets streamed to S3. We'll also want to impose a file size restriction.

I'm good with this PR as-is, and adding these as tickets. Up to you.

jjidt:
cool, yeah, let's take care of those in separate tickets, I think the document updatedAt pr will have conflicts with this one, so gonna go ahead and merge

dheniges (approved):
woooooo

PR #194, opened 2019-08-26
Adding spinners

jcontonio:
* Fixing issues where things could take too long (sending email) by adding spinners
* Adds spinner UI state so we can easily control when spinners should show
* Changes User Admin panel to be consistent with the back link
* Fixes re-invite bug (wasn't being called)


jjidt (changes requested):
code looks good, but seems like there's an issue with the spinner state, I'm getting permanent spinners on the users when the admin panel loads

Screen Shot 2019-08-28 at 9 26 35 AM

jjidt (approved):
👍

PR #195, opened 2019-08-27
allow image field creation without file

jjidt:


jcontonio (approved):
👍 Can create image field

PR #196, opened 2019-08-27
image size limit

jjidt:
- image size limit
- multer error handling
- temporarily write files to disk


jcontonio:
Changed the constant to `50`, but I can still update a field with a 1.5MB image

jcontonio (approved):
I get the file too large message 👍

PR #197, opened 2019-08-27
🎉 Drag and Drop Party 🎉

jcontonio:
* Adds new Photo field, with image preview, drag and drop
* Adds field moving prep - drag a field to another document (logs relevant info)
* Adds dragging and dropping of markdown files to the document list of a project, creates them with a body
* Adds body field to new document by default
* Focuses the last line of text in the document automatically
* Updates styling of the editor more

**Needs**
Update endpoint for the image field... images are not saving yet.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -5554,7 +5554,8 @@
           "version": "2.1.1",
           "resolved": false,
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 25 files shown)


jcontonio:
@jjidt when you get a sec, can you take a look at this? On line #22 of DocumentsListContainer.js, I can create a text field with a ` ` value, but not a `` value. Any way around that? 

I basically want to create an empty body field, so the placeholder appears in the textarea

jjidt:
@jcontonio check if the changes I pushed up fix the problems you were having

jcontonio:
It uploads 👍 thanks!
Working on the state change though, I had made some changes locally

jjidt (dismissed):
👍 Works great in different browsers, great to see this coming together.  Also since the timeout hack is for UI weirdness I'm cool with it.

jjidt (approved):

PR #201, opened 2019-08-30
implement hash id namespacing for s3 content uploads

jjidt:


api/src/integrators/s3.js
```diff
@@ -1,19 +1,23 @@
 import AWS from 'aws-sdk'
 import ono from 'ono'
 
-const { S3_IMAGE_BUCKET, AWS_SES_REGION } = process.env
+import { getHashIdFromOrgId } from '../libs/hashId'
+
+const { S3_CONTENT_BUCKET, AWS_SES_REGION } = process.env
```
dheniges, line 6: Need to add `S3_CONTENT_BUCKET` to example.env and update the 1pass .env file with these. Also noticed the `STRIPE_SECRET` isn't in the example.env while you're in there

dheniges (changes requested):
Looks good, just updating stuff with env vars per comment

dheniges (approved):
#greatsuccess

PR #202, opened 2019-08-30
Removed search field for now

jcontonio:
Removes search because we don't have it... just more prep for making this ready for users


jjidt (approved):
👍

PR #203, opened 2019-08-30
Empty states

jcontonio:
Adds empty states when there are no projects, and no documents

~*Need to remove logging, as we're logging for a different bug atm*~


jjidt (approved):
These look great 👍 Nice work

PR #208, opened 2019-09-09
Adding field name for image

jcontonio:
* Adds UI for updating image name in settings view
* Needs the ability to update name OR value (file) on an Image field


dheniges (dismissed):
There's some weirdness with the UI state here, not sure what's triggering what.
1. If you open the settings on an image, you can't double click to select all the text (but you can cmd + a)
2. The image gets a spinner on it when editing the text, if you haven't pressed update

Also in terms of naming, I missed this getting named "photo." "Image" seems like it captures all use cases better?

jjidt:
updated to fix image renaming

jcontonio:
This is ready for review

jjidt (approved):

PR #210, opened 2019-09-09
user/org avatar endpoints

jjidt:


api/src/controllers/users.js
```diff
@@ -11,6 +12,9 @@ import perms from '../libs/middleware/reqPermissionsMiddleware'
 import ACTIONS from '../constants/actions'
 import { requiredFields } from './payloadSchemas/helpers'
 import userSchema from './payloadSchemas/user'
+import avatarCoordinator from '../coordinators/avatar'
+
+const MAX_AVATAR_FILE_SIZE = 1024 * 1000
```
dheniges, line 17: I was going to say this is defined twice (org and user avatars) but thinking on it I don't know that they need to be tied together. So I think it's fine.

dheniges (commented):
What do you think we should do if there's an existing avatar on a user or org? Thinking through options:
1) Leave it, because we want to let them revert to previous avatars eventually?
2) Delete the old avatar file after the new one successfully saves
3) ??? Profit!

dheniges (approved):
Approving. Add deletion in a new PR so we can get the test fixes

PR #215, opened 2019-09-11
user and org avatar deletion

jjidt:


jcontonio (approved):
Code looks good to me 👍

PR #217, opened 2019-09-13
Form error/success/submitting tracking

jcontonio:
* Tracks which form is doing what 
* Displays a success icon if the form was submitted successfully
* Displays an error message if the form had errors
* Displays a spinner if the form is submitting
* Adds max image size on avatars

**To test**

I commented out some client-side validation for testing purposes

1. Go to User Profile
2. Submit the "Your information" form, deleting the user's name (bad data)
3. Submit the avatar form with a successful image
4. You should see errors for the first one, success for the second one

Rinse repeat for http://localhost:3000/d/admin/organization


web/src/client/js/actions/user.js
```diff
@@ -57,31 +57,41 @@ export const createUser = (values) => {
   }
 }
 
-export const updateUser = (userId, body) => {
```
jjidt, line 60: we should probably make the formId the last param everywhere, so that it could be optional, or if we end up passing addition UI-related ids we can put them all in an options object

jjidt (commented):
I like how this is set up. It seems like it's still not handling 400s correctly though coming back from the server, or maybe that's still something you need to set up?

jcontonio:
It handles the 400 error if you submit with a blank name. Is there something different happening for you?

jjidt:
if you set the file size limit on the client to a higher value than the server and upload a large file, it'll come back with a 400 and not display it. Maybe the server error data needs to be configured in a better way.

jcontonio:
Ahh ok. Let’s figure out when you get a moment next week

jjidt:
@jcontonio errorDetails always coming back now, tie it in and I'll give it a once over and we can get this merged

jcontonio:
@jjidt looks good, however getting "invalid file" for the 413 instead of "file too large" or whatever it should be. Not sure if 413 got swapped with 415. 

But you can now try uploading a larger file and get the validation message in the UI 👍

jjidt:
fixed

jcontonio:
Looking good!

image

jjidt (approved):

PR #221, opened 2019-09-18
Fixing size of brand area

jcontonio:
The current Portway logo was messed up and I didn't catch it because I had the bonkeybong logo in there


jcontonio:
Closing this because it's different now with the image cropper

PR #222, opened 2019-09-18
Adding image cropper to avatar uploading

jcontonio:
* Adds ImageCropperComponent that returns a cropped image blob and blob URL
* We can submit the blob as FormData, or render the image in the dom


jjidt (approved):
this is really nice 👍

PR #226, opened 2019-10-01
Copying and moving fields

jcontonio:
Can copy/move fields to different documents


jjidt (approved):
code looks good, works great! 👍

PR #227, opened 2019-10-02
password change

jjidt:
Password change, API and Web

@jcontonio wanna take a look at this first and see if you want to do anything in a different way?


jcontonio:
Yep, will take a look @jjidt

jcontonio:
image

We need errorDetails populated like all other validation things, so we can display a message with the field. I think something like password, newPassword, confirmPassword

jcontonio:
btw I updated the branch to get the form to work with validation

jjidt:
@jcontonio passing this back to add field validation

jcontonio:
@jjidt ready for ya

jcontonio (approved):

PR #228, opened 2019-10-05
Digital Ocean dev kube environment

dheniges:
- Adds base route to API for ssl validation with letsencrypt
- Adds DO k8s config files
- Does other things I probably forgot


jjidt (approved):
👍 Looks good, everything still works locally

PR #230, opened 2019-10-07
Lazy loading code & extreme bundle caching

jcontonio:
Added a new thing in `web`. You can run the local production build by going in there and running `./run.sh local`


web/src/client/js/components/Admin/AdminDashboardComponent.js
```diff
@@ -13,8 +13,10 @@ import { Panel, PanelNavigation, PanelContent } from 'Components/Panel'
 import OrgPermission from 'Components/Permission/OrgPermission'
 import AdminUsersContainer from './AdminUsers/AdminUsersContainer'
 import AdminUserViewContainer from './AdminUserView/AdminUserViewContainer'
-import AdminBillingContainer from './AdminBilling/AdminBillingContainer'
-import AdminOrganizationContainer from './AdminOrganization/AdminOrganizationContainer'
+
+// Lazy loading these because most users won't see these sections
+const AdminBillingContainer = lazy(() => import(/* webpackChunkName: 'AdminBillingContainer' */ './AdminBilling/AdminBillingContainer'))
```
dheniges, line 18: remove comments? Looks like the chunk names are the same as the imported name (which presumably it would have to be)

web/src/client/js/components/Admin/AdminUsers/AdminUsersCreateForm.js
```diff
@@ -36,7 +37,7 @@ const AdminUsersCreateForm = ({ cancelHandler, errors, submitHandler }) => {
           id="userName"
           label="Full name"
           name="name"
-          onChange={(e) => debounce(200, setName(e.target.value))}
+          onChange={e => debounce(200, setName(e.target.value))}
```
dheniges, line 40: I thought parens were required in our lint rules?

web/src/client/js/components/ProjectSettings/ProjectSettingsComponent.js
```diff
@@ -1,13 +1,14 @@
...
 import PropTypes from 'prop-types'
 import { Link, NavLink, Redirect } from 'react-router-dom'
 
 import { PATH_PROJECT } from 'Shared/constants'
 import { Panel, PanelNavigation, PanelContent } from 'Components/Panel'
 import ProjectSettingsInfoContainer from './ProjectSettingsInfo/ProjectSettingsInfoContainer'
-import ProjectSettingsTeamsContainer from './ProjectSettingsTeams/ProjectSettingsTeamsContainer'
 import ProjectSettingsTokensContainer from './ProjectSettingsTokens/ProjectSettingsTokensContainer'
 
+const ProjectSettingsTeamsContainer = lazy(() => import(/* webpackChunkName: 'ProjectSettingsTeamsContainer' */ './ProjectSettingsTeams/ProjectSettingsTeamsContainer'))
```
dheniges, line 10: does webpack require these comments?

web/src/server/app.js
```diff
@@ -40,8 +40,8 @@ app.use(passport.initialize())
 if (devMode) {
   app.use(express.static(join(__dirname, './public')))
 } else {
-  const oneDay = 86400000
-  app.use(gzipStatic(join(__dirname, './public'), { maxAge: oneDay }))
+  const expire = 3.154e+10 // 1 year - Google suggests this
```
dheniges, line 43: I think this is the case, but to confirm there are no images or anything except webpack created and versioned bundles of css and js in here, right?

dheniges (commented):
It works. But loading the app is literally over 100 javascript files and this doesn't include http2 support. It looks like most browsers cap concurrent connections at 17. So while the payloads are small, this has to be a huge load hit, especially for higher latency connections with that many ssl handshakes. It seems like this needs http2 support before we merge it. Thoughts?

dheniges:
Update from slack convo: to alleviate http1.1 issues there are webpack plugins that can help group bundles. Getting the number of initial js bundles loaded to ~10 should be good.

web/src/client/js/components/Admin/AdminDashboardComponent.js
jcontonio: Those comments are magic webpack comments. They need to be there so we can get named chunks.

web/src/client/js/components/Admin/AdminUsers/AdminUsersCreateForm.js
jcontonio: Our rules complained that because it's returning a function that no parens should be there

web/src/client/js/components/ProjectSettings/ProjectSettingsComponent.js
jcontonio: It doesn't require them, but there's no way to identify the chunk without them. Webpack will number them instead. I like to have the names so we can look at the stats.

web/src/server/app.js
jcontonio: Yep. Everything is hashed with webpack. We barely have any images anymore, maybe just our logo and social pngs for meta tags. Every icon in the app is a React component.

jcontonio:
k @dheniges this is ready for a re-review. Reduced the files to 10, imported the two svg's we're using explicitly (I will do the social pngs in the other branch). That'll solve the caching issue for those handful of files. 

We could use maxSize here to split the app bundle, but it's not executing. Wonder if when we switch to HTMLWebpackPlugin (other branch) if that is still an issue. Left everything commented so we remember what is what.

dheniges (approved):
Nice work on all this 👏 

Dev still seems good, hot reloading etc., and the prod build looks good too.

PR #232, opened 2019-10-09
Initial render work

jcontonio:
This is the start of an initial skeleton render with critical CSS. I still need to work out the CSS and skeleton HTML but am going to wait for another PR.

We won't achieve perfection, as we have to wait for HTMLWebpackPlugin 4.0 to be out of beta, and for the inline styles plugin to catch up. We'll be in a really good spot after that though.

## What this does

This changes the Express controller for our React app. It loads a static HTML that is generated by webpack instead of an EJS template. There are adjustments for dev vs prod in Express and our webpackDevMiddleware library.

You should see a skeleton of HTML that is styled, so that on a slow connection we aren't looking at a blank page for 11-15 seconds while our vendor code downloads. Once React is ready the skeleton will get wiped out and overwritten with our app.

## What it doesn't do yet

initialRender.scss should be inlined to a <style> tag in the static HTML template. We're not getting that yet due to webpack sucking.


jcontonio:
Can anyone take a look and see why Express's sendFile is throwing this error back, but still totally working?

```
GET /Users/jcontonio/Sites/portway/web/dist/server/app.html?_sw-precache=59d8ea114dff6c3ca348a1bcf37495d9 404 2.711 ms - 136
GET /Users/jcontonio/Sites/portway/web/dist/server/app.html.gz?_sw-precache=134c445c43e8272598a671af464d273f 404 2.297 ms - 136
```

In controllers/app. app.html is generated from webpack. It all works in the browser, but a weird 404 in the logs

PR #233, opened 2019-10-10
updating org plan

jjidt:
before testing, you need to set up a product in your stripe test environment (make sure the `view test data` toggle is set in the stripe dashboard) and add 2 plans to the product with defined IDs `SINGLE_USER` and `MULTI_USER`


jjidt:
@jcontonio This is ready for you to:

- hook up multi/single highlighted selection to the organization.plan field
- prevent plan selection until the user has billing information
- do what you need to do with organization.subscriptionStatus

jcontonio:
@jjidt back at you - things to check:

* Header doesn't have UPGRADE if org has a subscriptionStatus !== trialing or null
* Selecting a plan should enable the submit button, clicking it should be like normal forms with progress spinner and a success checkmark
* When in single-user mode, Billing seats, Project users, Admin users, Project settings users should all not be accessible
* Price should reflect the plan (minus the seats at the moment)

api/src/db/migrations/20190221184519-create-organization.js
```diff
@@ -26,6 +26,12 @@ module.exports = {
       avatar: {
         type: Sequelize.STRING
       },
+      plan: {
+        type: Sequelize.STRING
+      },
+      subscriptionStatus: {
+        type: Sequelize.STRING
+      },
```
dheniges, line 34: Going forward, since we have a deployed environment, let's start adding new migrations

web/src/client/js/components/Admin/AdminPayment/AdminPaymentComponent.js
```diff
@@ -63,7 +72,7 @@ const AdminPaymentComponent = ({ errors, isStripeOpen, isSubmitting, openStripeH
       <div className="admin-payment">
         {!isStripeOpen && orgBilling && orgBilling.source.brand &&
         <>
-          <p>You will be billed <b>$10</b> on <b>September 19, 2019</b></p>
+          <p>You will be billed <b>${cost}</b> on <b>September 19, 2019</b></p>
```
dheniges, line 75: How do we want to track placeholders like this date in the app now? Asana backlog or TODO?

web/src/shared/constants.js
```diff
@@ -6,8 +6,14 @@ const FIELD_TYPES = {
 }
 
 const PLAN_TYPES = {
-  SINGLE: 'single',
-  MULTIPLE: 'multiple'
+  MULTI_USER: 'MULTI_USER',
+  SINGLE_USER: 'SINGLE_USER',
+  TRIAL: 'trialing',
```
dheniges, line 11: Where does the subscriptionStatus get set to "trialing"? Is that from Stripe? I don't see this value getting set anywhere, just read. A comment would be useful here

dheniges (commented):
I know this is a work in progress, but not sure what's being tracked, so comments:

Bugs/issues:
- Highlight color over plan selection turns on when you mouse out of the area. Flashes oddly.
- Can attempt to update plan without billing info, but no error message shows up in UI. Although with a new signup (non-seed data), it returns a generic 500 and shows conflicting error messages in the UI ("See error in form" but form has no errors, and the 500 popover message)

dheniges:
Oops, forgot to mention in my review, but part of the review. We need docs on the Stripe setup in the docs folder.

It should include:
- Info on what Stripe features we're using
- How products and plans are setup, and steps for setting it up

jcontonio:
> I know this is a work in progress, but not sure what's being tracked, so comments:
> 
> Bugs/issues:
> 
> * Highlight color over plan selection turns on when you mouse out of the area. Flashes oddly.
> * Can attempt to update plan without billing info, but no error message shows up in UI. Although with a new signup (non-seed data), it returns a generic 500 and shows conflicting error messages in the UI ("See error in form" but form has no errors, and the 500 popover message)

There's a ticket to handle the plan selector separately for this

web/src/client/js/components/Admin/AdminPayment/AdminPaymentComponent.js
jcontonio: This is getting worked on in JJ's current work, but yeah, I think we should have maybe a "Before release" ticket with subtasks for all the things we find?

web/src/shared/constants.js
jcontonio: "trialing" gets set on new org creation. Right now you don't see it because our seed data is just null

PR #236, opened 2019-10-11
Adding multiple themes

jcontonio:
Open up `head.ejs` in web/src/server/views/partials and change the body class to one of the following:

`theme--default`
`theme--salmon`
`theme--yellow`

Check out the sign in, register, log in, and the app


jcontonio:
Ok, dark mode is updated. Looks cool with the new themes.

dheniges (dismissed):
:+1: has colors

dheniges (approved):
:+1: still has colors

PR #237, opened 2019-10-14
Adding password meter

jcontonio:
* Adds client-side password checking
* Adds new Popover UI component
* Only loads zxcvbn when UserSecurity component gets loaded, so we don't have a giant library in the regular endpoint


jjidt (changes requested):
Code looks good, there's some overlap that can happen with the dual popups

Screen Shot 2019-10-14 at 4 53 32 PM

jjidt:
Also, go through the whole sign up flow, I wasn't seeing these for the initial password page

jcontonio:
@jjidt pull the latest and all of those problems should be solved

jjidt (approved):
👍

PR #239, opened 2019-10-16
Trial period UI

jcontonio:
* Adds in UI for the billing component and header component when we are in a Trial or Locked status
* Adds in UI for Locked view everywhere but the admin section 
* Updates a few colors and switched to the Salmon theme


web/src/shared/constants.js
```diff
@@ -8,9 +8,28 @@ const FIELD_TYPES = {
 const PLAN_TYPES = {
   MULTI_USER: 'MULTI_USER',
   SINGLE_USER: 'SINGLE_USER',
-  TRIAL: 'trialing',
 }
 
+// https://stripe.com/docs/billing/lifecycle#subscription-states
```
dheniges, line 13: 👏

dheniges (changes requested):
I didn't visually see all the states, but combed over the code. Only issue that needs updating is the gray on gray of the "Upgrade" button:

Screen Shot 2019-10-18 at 10 19 39 AM

dheniges:
Another comment: even after you add payment info, the trial message still says to update your payment info in order to pick a plan. Not sure if we have a flow in mind for adding card info and selecting the plan?

jcontonio:
That sounds like something isn't working correctly, or the org isn't getting refetched. Once you add payment info the org should turn to 'active' or 'invalid', etc.

That gray on gray is weird, that may be another state you're in that wasn't checked. Will look into it.

jcontonio:
It's not an org refetch problem. I think this will get fixed with @jjidt's hooks coming back from Stripe. I think that's in another PR. Once the payment is added, Stripe needs to contact our server and adjust the subscription (me thinks). This should get merged now though, functionality is right, and styling is updated.

dheniges (approved):
:+1:

PR #240, opened 2019-10-16
subscription info

jjidt:
- provides subscription info in billing request
- camel case all billing fields returned from portway api


jcontonio (approved):
Looks good 👍 I added the day in there on the nextBilling date format. 
Cost doesn't update when switching plans, but does with a page refresh. Dunno if there's a nice way to deal with that or if we have to hit the org endpoint again?

PR #242, opened 2019-10-17
CSS Audit / Colors / Type

jcontonio:
* Moved to css variables instead of sass variables across the whole project
* Switched it all to the salmon theme by default, fixing any color changes from the new color palette
* Added in new Domine serif and updated all type size according to Claire's doc
* Updated editor type styles according to Claire's doc
* Added in new logo


dheniges:
It looks good from a clicking around the app perspective. Looked at most pages in both Chrome and FF.

Only question: there's some commented out css in places. Do you want that in there? Up to you, since you're the only one that does css.

jcontonio:
@dheniges yep, I want commented out CSS in there. It's to let me know "these classes exist", they just don't have styling on them.

dheniges (approved):
purdy colors 🌈 🏳️‍🌈 :salmon:

PR #245, opened 2019-10-18
Feature pagination

jcontonio:
Opening this because we forgot to


dheniges (commented):
Only a bit of work here :)

We should all click around on this. @jcontonio I commented out some scss vars that probably need replacing with css vars. See Paginator.scss

I tried to go through all the functionality. User search works, as does pagination and sorting, and state is preserved in the url

jcontonio:
Clicked through, working for me 👍

dheniges (approved):
Jay says it's cool

PR #248, opened 2019-10-18
plans/seats

jjidt:
@jcontonio the current subscription seat count will be available in the client on subscription.currentSeats


api/src/coordinators/billing.js
```diff
@@ -73,8 +74,40 @@ const subscribeOrgToPlan = async function(planId, orgId) {
...
+      const message = `Cannot have less than ${MULTI_USER_DEFAULT_SEAT_COUNT} seats on a multi user plan`
+      throw ono({ code: 409, publicMessage: message }, message)
+    }
+    updateSeats = seats
+  }
+
+  if (updatePlanId === PLANS.SINGLE_USER && currentPlanId === PLANS.MULTI_USER) {
+    const message = 'Cannot downgrade from multi user to single user plan'
+    throw ono({ code: 409, publicMessage: message }, message)
+  }
```
dheniges: This if/else block concerns me a bit, as there are lots of cases being handled based on a combination of loosely coupled variables. The API endpoint seems reasonable at first glance (takes a `seats` number and a `planId`), but the combination of business logic between plans and seats and handling that in this `if/else` block is dense. The test coverage looks good though. Looking at the code, I wonder if we should separate out create vs. update, as both here and the stripe integrator are pulling out create vs. update and running separate logic. Potentially, this could be a `createOrUpdate` function since they both return a subscription, and then immediately branch off to a create coordinator function and an update coordinator function. Same for the stripe integrator. What do you think?

web/src/client/js/components/Admin/AdminPayment/AdminPaymentComponent.js
```diff
@@ -30,7 +29,7 @@ function renderCardLogo(brand) {
 
 function getTotalCost(subscription) {
   // TODO calculate total cost from flatCost + seatCount*costPerSeat
-  return subscription.flatCost
+  return subscription.flatCost || '000'
```
dheniges: Is this covering the scenario where flatCost may be undefined because the user is on a trial? I think it's worth a code comment

dheniges (changes requested):
See my comments on the subscription re: "plan v. seat" and "create v. update". My hunch is we can break things down a bit to make it simpler to follow and update in the future.

api/src/coordinators/billing.test-u.js
```diff
@@ -32,21 +54,211 @@ describe('billing coordinator', () => {
...
+    describe('when no stripe customer is found', () => {
+      it('should throw an error with status code 409 ', async () => {
+        BusinessOrganization.findById.mockReturnValueOnce({ stripeId })
+        stripeIntegrator.getCustomer.mockImplementationOnce(() => {})
+        await expect(billingCoordinator.subscribeOrgToPlan(PLANS.MULTI_USER, orgId))
+          .rejects.toEqual(expect.objectContaining({ code: 409 }))
+      })
+    })
+
+    // describe('when trying to change seat count on a single user plan', () => {
```
dheniges: you still want these two commented tests?

dheniges (commented):
Comment on commented tests, otherwise it looks excellent :+1:

jjidt:
@dheniges think we're ready to merge this when you give it the go ahead

dheniges (approved):
:tada:

PR #250, opened 2019-10-19
payment subscription

jjidt:
- automatically upgrades user to a paid plan when they put in credit card info
- re-subscribes if they fix an inactive card
- we'll want to play around with this to make sure we keep billing dates etc.
- passes null values for all undefined fields for org billing


jjidt:
TODO: send null values for empty subscription fields

dheniges (approved):
Good comments, still UI things to iron out and probably more on subscriptions but everything in here looks good

PR #246, opened 2019-10-18
CI/CD deployments

dheniges:
Requires #241 

- See docs in `README.md` and `docs/pipelines.md`
- Adds support for deploying tagged releases to the dev environment on Digital Ocean
- Updates test and release pipelines to make use of parallelism 💪 🍄 

This requires a revised workflow. This can be reviewed for code, then we'll merge around some branches:
1. Merge this PR to master
1. Merge master into `release` and push to origin
1. Move any other open PRs to point to `release`

PR #251, opened 2019-10-19
Add migrate job to helm chart

dheniges:
As a pre-upgrade step, the helm chart for portway will run a Kubernetes job that pulls the API image and runs a database migration! The chart will not deploy if the job does not run successfully

Requires #246


jcontonio (approved):
👍

PR #253, opened 2019-10-22
Fix webpack build + ability to update payment info

dheniges:
Allows use of the new db migration step in the helm chart for deployed environments


jcontonio:
I don't know if this is related to this PR, but I can't update my payment information successfully now. It errors out the client here, in AdminPaymentComponent

`const cost = orgBilling.subscription && toCurrencyString(getTotalCost(orgBilling.subscription)) // @todo update with seats`

jcontonio (dismissed):
Cleaned up AdminUsersComponent so that webpack can build it
Cleaned up AdminBilling so that it won't error out when adding payment information

dheniges:
I'm taking this approval for the migrations into #255 so your updates can be reviewed separately. Thanks for testing the migration stuff

dheniges:
how are you testing the Admin users section? I can't get to it because I can't select a plan

jcontonio:
I wasn't testing getting to it, it was failing compile in webpack so you couldn't get web to work at all. So that work fixed some jank that the paginator PR brought in that broke the component from compiling

dheniges:
:stamp_of_approval: because I can't officially review it

jcontonio (approved):
I approve my changes

PR #254, opened 2019-10-23
Remove webpack API url build arg dependency

dheniges:
Why have a build arg when you can party with ejs?

This allows us to create a release that's deployable on any environment.


jcontonio (approved):
👍 adds API URL here

PR #256, opened 2019-10-29
FormField & Color consistency

jcontonio:
* Adds new "theme" to better manage dark/light themes
* Fixes iPad issues with hovers
* Refactors all form fields into one FormField that can be multiple types
* Adjusts ProjectSettings form to use the new Form / FormField /  Status pattern


dheniges (approved):
more colors than before :+1:

PR #258, opened 2019-10-30
Responsive UI start

jcontonio:
* Adds responsive panel system
* Adds responsive document / document list


jcontonio:
Going into feature-plans because it requires some newer stuff

dheniges (approved):
It responds :+1:

PR #259, opened 2019-11-05
Forms - Not so minor updates

jcontonio:
* Makes form markup more consistent
* Comments css in forms.scss, and removes a bunch of unneeded things
* Removes Bulma buttons from marketing site theme

To test:
You shouldn't notice a damn thing 🕵️‍♀️


dheniges (changes requested):
the user icon isn't displaying, maybe related to this?
Screen Shot 2019-11-05 at 10 12 59 AM

dheniges (approved):

PR #260, opened 2019-11-05
Disables animation for people with the setting on

jcontonio:
To test:

1. Go to System Preferences -> Accessibility -> Display and turn Reduced Motion: on
2. See that nothing animates. The most noticeable is the ContentMenu in a document, its "+" sign rotates to an "x" if the setting is off, or responsively a document loading


dheniges (approved):
that's pretty nifty

PR #263, opened 2019-11-08
worker thread to run markdown parsing on text fields

dheniges:
- Utilizes node.js `worker_thread` module
- Offloads structured json parsing of raw markdown so as not to block main thread
- Good tests

Requires #261


dheniges:
A note on performance with this: https://docs.google.com/spreadsheets/d/1Ioxaxzxeh2HUAk-2L0qMhyLPXlSOvdWw6BoZQSWBplE/edit#gid=0

Decision for now is to leave it running workers, but it's not as useful as I had hoped

jjidt (changes requested):
getting a 500 and

` Error: Input data should be a String` from MarkdownIt

when trying to create a new body in the client app

dheniges:
@jjidt updated to fix that bug + added more tests in both the processing lib and the coordinator

jjidt (approved):
👍 looks good

PR #264, opened 2019-11-09
Plan Confirmation

jcontonio:
* Disables downgrading and notifies user
* Confirms on plan change
* Fixes form field width on credit card zip and state


dheniges:
The UI looks good, but I can click on whatever plan I'm currently on and it enables the "Update Plan" button which I can then click and get the payment confirmation

jcontonio:
Fixed that up @dheniges

dheniges (approved):
:+1:

PR #266, opened 2019-11-12
Adding seat management

jcontonio:
* Adds Seat management in admin section
* Adds cancelHandler to Form component

image

@jjidt having trouble with the seats endpoint / result. 

1. Need `errorDetails` populated like all the other endpoints
2. I am setting { seats: 6 } for example, and I am not getting any data back to update the UI, nor is the organization subscription `includedSeats` field getting updated on a hard refresh
3. Right now I have 6 out of 5 seats (`currentSeats` out of `includedSeats`).

I also have a ton of users, so the user count !== currentSeats at the moment


jjidt (approved):
👍 works great, I went ahead and added the user count check to the api route as well with accompanying tests

we'll want to test out a bunch of scenarios because the invoices seem to get out of whack with our stripe settings, but there's nothing here that would be causing that

Screen Shot 2019-11-14 at 1 23 42 PM

jcontonio:
Thanks @jjidt!

PR #267, opened 2019-11-12
Use Cloudfront CDN for S3 assets

dheniges:
- Adds the `CDN_HOSTNAME` environment variable
- `CDN_HOSTNAME` plugged into all the kubes (via helm)
- S3 integrator expanded to return the cloudfront equivalent url. This uses the node url lib, so I think it's ok in an integrator?


jjidt (approved):
👍

PR #268, opened 2019-11-12
stripe hooks

jjidt:
- add stripe hooks for payment processing that trigger emails to the billed user
- after adding the STRIPE_HOOK_SECRET to your local .env file, you should be able to add a new user with your bonkeybong.com email address, enter credit card info, and get a payment successful email
- I think we'll want to update org subscription status and any other cached values through a daily job rather than webhooks, it should be possible to deactivate an account after a certain number of payment failures


dheniges:
The rest of the app works, but I am not setup to test the webhook part. Code looks fine, and if you've confirmed the webhook functions as expected JJ I'm good merging this

jjidt:
webhooks work, but I can walk you through a demo before merging

dheniges (approved):
Looks good to merge 👍

PR #269, opened 2019-11-12
Fixing password validation on reg

jcontonio:
* Fixes validation steps on password fields
* Allows you to have a bad password, but warns you
* Keeps the minimum to 8 chars


dheniges:
I can't seem to get a screenshot, but if I start typing in the confirmation password and then unselect it, I get the password confirmation error message as expected. However, I cannot click the 1st password field, it automatically refocuses on the password confirmation field with the error message.

So if I've mistyped the original password, or decide to change it, I can't

dheniges (changes requested):
It's still possible to get into broken states in this form without doing anything crazy. Switching back and forth between fields can cause validations that should show up to disappear, or to flash. I think it's related to the setTimeout of 1 second that wipes out validation messages for both fields. Most concerning I can have matching passwords but the submit button is still disabled.

I guess react is helping render here, sort of, but I think direct DOM manipulation or having react control the form fields and state is going to be easiest. I think separating out the logic between the password strength validation and the password matching validation, including show/hide, will also make it easier to work with and reason through.

jcontonio:
@dheniges broke out the UserSecurityFields into a React component that is being shared between the registration form and the password change form.

Also:

* I added a `disabled` property to the Form component so we can control that better
* I think I have all cases down in the fields for what someone could do

dheniges:
I'm afraid you might kill me. Buuuttt, this is nearly perfect. Except if you type a password it doesn't like, and then type it again in the confirmation, it tells you they don't match, even if they do:
Screen Shot 2019-11-14 at 5 04 58 PM

jcontonio:
I *think* this is good to go...

jcontonio:
FYI, this allows you to have a bad password... 

image

dheniges (approved):
🌮 🎉 🎋 💯 🎖 🏅  It works great, can't find anything wrong with it 👏 
Nice work on finagling all those validations together, and using it on both forms.

I pushed one minor tweak to add `name="token"` to the hidden input field because registration wasn't working. Backlogged an item to add a "Password change success" message when a user changes the password in the app.

jcontonio:
@dheniges this does the green checkmark, so I am going to merge it. We'll do a flash thing in another ticket.

PR #271, opened 2019-11-13
Updating icons

jcontonio:
* Updates icons to new outlined versions
* Adds arrow icon with direction options

To test just click around and look for messed up icons


dheniges (approved):
Didn't notice anything out of line

PR #274, opened 2019-11-14
org status on payment.failed hook

jjidt:


jjidt:
- updates cached subscription status on payment failed and payment succeeded events
- prevents access to content endpoints if user isn't `active` or `trialing`
- logs warning if a stripe customer ends up with multiple subscriptions
- ends trial when user puts in card info (easiest solution for now @dheniges) we'll probably want our own subscriptionStatus values for the org that don't come directly from stripe, something like 'PENDING' for users who are still in trial, SINGLE_USER plans, but have already put in card info

dheniges (approved):
Looks good, tested non-hook stuffs without issue

PR #275, opened 2019-11-14
use dynamic API url to populate example token endpoints

dheniges:
Now the project tokens page will show whatever API it's connected to


jcontonio (approved):
Works nice

PR #280, opened 2019-11-15
Reconfigure controller middleware loading

dheniges:
- Ensure body parsing happens first
- Makes sure various options for each controller don't override each other (modify the same middleware array rather than replacing)

### DIFF
--- a/api/src/controllers/loader.js
+++ b/api/src/controllers/loader.js
@@ -68,18 +68,18 @@ const loadControllers = (router, controllers, middleware = [], routerOptions) =>
     const controllerRouter = express.Router(routerOptions)
     controller(controllerRouter)
 
-    let expandedMiddleware
-    // activate body parsing middleware for each route, default to json parsing, unless otherwise specified
+    const expandedMiddleware = [...middleware]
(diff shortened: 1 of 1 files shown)


api/src/controllers/loader.js
```diff
@@ -68,18 +68,18 @@ const loadControllers = (router, controllers, middleware = [], routerOptions) =>
...
-      expandedMiddleware = [...middleware, bodyParser.json()]
+      expandedMiddleware.unshift(bodyParser.json())
     }
-    // activate subscription status check middleware for each route, default is to only allow active orgs access
-    // unless allowInactiveOrgAccess is set to true
-
+    // activate subscription status check middleware for each route, default is to only allow active
+    // orgs access unless allowInactiveOrgAccess is set to true
     if (!controllers[path].allowInactiveOrgAccess) {
-      expandedMiddleware = [...middleware, checkActiveOrgStatus]
```
dheniges, line 82: @jjidt this was overwriting expandedMiddleware added from deleted line 74 or 76, so some endpoints didn't have any body parsing

jjidt (approved):
👍 fixed it

PR #282, opened 2019-11-15
Editor - Move document menu to document list

jcontonio:
* Moves document menu out of the document header
* Tightens the document list down
* Changes highlight text to yellow

image


web/src/client/js/components/Header/HeaderComponent.js
```diff
@@ -28,7 +28,7 @@ const renderBrandLogo = (logo) => {
 const HeaderComponent = ({ brand, isFullScreen, section, subscriptionStatus }) => {
   const upgradePillClasses = cx({
     'pill': true,
-    'pill--red': LOCKED_ACCOUNT_STATUSES.includes(subscriptionStatus),
+    'pill--red': !LOCKED_ACCOUNT_STATUSES.includes(subscriptionStatus),
     'pill--gray': subscriptionStatus === SUBSCRIPTION_STATUS.TRIALING || subscriptionStatus === null,
```
dheniges: This is assigning both `pill--gray` and `pill--red`. Is red not supposed to be if it's locked?

dheniges (commented):
Everything works that I tried. Thanks for the smaller PR. Only question about what pill class is supposed to be given to what.

web/src/client/js/components/Header/HeaderComponent.js
jcontonio: This is ready for review @dheniges, removed the debugging thing

dheniges (approved):
Pills look good now

PR #290, opened 2019-11-19
unpublish doc

jjidt:


api/src/controllers/documents.js
```diff
@@ -38,6 +39,8 @@ const getDocument = async function(req, res, next) {
   }
 }
 
+//TODO: should probably move publish and unpublish to projectDocuments controller and save document lookup,
```
dheniges, line 42: Maybe, but caching will save us :) Also I much prefer to look up the document twice and not have the coordinator take a document sequelize object but just ids.

dheniges (approved):
No more docs!

PR #291, opened 2019-11-20
Editor - New toolbars

jcontonio:
* Separates Project and Document toolbars into two different components
* Refactor the layout a bit of the sidebar / content area
* Removes publish stuff from document header and moves to footer
* Moves content menu into the document footer

Still a bunch more to do

Screen Shot 2019-11-19 at 5 56 09 PM


dheniges (approved):
Looking good, couldn't break anything

PR #292, opened 2019-11-20
fixing bad logic in JWT project token auth

dheniges:
- Bonus: adds explicit 404 error to `organization.findById`
- Bugs in the JWT project token auth would not throw appropriate errors and would return a valid "token" with no orgId, id, etc. Fixed code + new tests ensure this is no longer the case


jjidt (approved):
👍 nice catch

PR #294, opened 2019-11-22
Editor - New Edit mode with better drag and drop

jcontonio:
PR'ing this so we can review just the code involved and see if we can make it any cleaner. Will do styling and more in future PRs

For Drag and Drop, the key thing to look at here is DocumentFieldsContainer


dheniges (approved):
It works, and I like the experience better than before.

It seems to work pretty flawlessly when dragging around the basic fields, like string or number. But when the text fields are involved it gets pretty weird. It's fairly easy to get the placeholder shadow area to jump back and forth super rapidly between the top and bottom of a text field even if you don't move the mouse. The scroll into view function is a bit hit or miss. When it works it seems pretty obvious that it's working, but I think coupled with whatever other drag issues with the content areas it can get wonky. I disabled it and the experience didn't get any better, so it may be other events interfering with it working, not sure.

jcontonio:
Cool! Yeah there’s some more to figure out there.

PR #295, opened 2019-11-26
Adding in unpublishing

jcontonio:
* Adds un-publish menu item in document menu
  * Un-publish is disabled when there is no published version
* Adjusts publish button:
  * if the document doesn't have a published version, button says "Publish document"
  * if the document has a published version, button says "Publish changes"
  * publish button is only enabled if there is no published version, or if updatedAt is >= lastPublishedData
* Sets the document lastPublishedAt to whatever the server sends back

### DIFF
--- a/web/src/client/js/components/DocumentsList/DocumentsListContainer.js
+++ b/web/src/client/js/components/DocumentsList/DocumentsListContainer.js
@@ -71,11 +71,11 @@ const DocumentsListContainer = ({
   function unpublishDocumentHandler(document) {
     const message = (
       <>
-        <p>Un-publish the document <span className="highlight">{document.name}</span>?</p>
+        <p>Unpublish the document <span className="highlight">{document.name}</span>?</p>
         <p>If you are using this document in a live application, it will be removed.</p>
       </>
     )
-    const confirmedLabel = `Yes, ub-publish this document`
+    const confirmedLabel = `Yes, unpublish this document`
     const confirmedAction = () => { unpublishDocument(document.id) }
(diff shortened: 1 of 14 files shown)


web/src/client/js/components/DocumentsList/DocumentsListContainer.js
```diff
@@ -67,9 +68,21 @@ const DocumentsListContainer = ({
...
   }
 
+  function unpublishDocumentHandler(document) {
+    const message = (
+      <>
+        <p>Un-publish the document <span className="highlight">{document.name}</span>?</p>
+        <p>If you are using this document in a live application, it will be removed.</p>
+      </>
+    )
+    const confirmedLabel = `Yes, ub-publish this document`
```
dheniges: "ub-publish". I would call it "Unpublish" without the hyphen.

dheniges (changes requested):
I would change the wording everywhere from "Un-publish" to "Unpublish." Hyphens are usually only used to avoid confusion with an existing word, and "unpublish" is not an existing word. Yet.

Functionality works as expected, but got me thinking on the UI. If I'm looking at a document, I would expect I could perform any action on that document. Especially in the mobile view, where I can't see the document list. This would be for another ticket, but what about clicking the settings "Gear" icon exposing "Unpublish" and "Delete" doc buttons somewhere?

jcontonio:
@dheniges we should definitely do that, good idea. I could bring in the document menu next to the gear icon if we're in the small menu

dheniges (approved):
:tada:

PR #296, opened 2019-11-26
Password change notification

jcontonio:
* Displays a successful notification when changing the password
* Resets the form correctly upon successful submission


dheniges (approved):
:+1:

PR #298, opened 2019-11-26
Logging!

dheniges:
- Adds capability to add various loggers, defaults to console
- Adds support for r7 insight logging
- Adds custom http request logging, including portway customizations (namely user at this point)
- Adds cuid library for generating a unique machine name for logging. Can/should also be used for future uniqueId needs

The express middleware has a bunch of logic in it, but it's highly dependent on the req/res objects, so not sure on automated testing


jjidt (approved):
👍 nice
don't think automated testing is necessary here

PR #299, opened 2019-11-27
Audit log to an audit destination

dheniges:
Requires #298 

Uses a separate logger instance for insight ops so we can group and track audit logs separately. They have an API we can use to download the logs and save them on S3. Opening a new ticket for that work.

Audit logging defaults to console if no audit log token provided.


api/src/integrators/audit.js
```diff
@@ -1,6 +1,19 @@
 // Controller audit log integrator
 // receives an object with namespace info
 // currently logging to the console
```
jjidt: I think we can update the comment here now

jjidt (changes requested):
Just updating the comment, everything looks good 👍

api/src/integrators/audit.js
dheniges: updated

jjidt (approved):
👍

PR #300, opened 2019-11-27
cancel subscription

jjidt:
- adds endpoint to cancel org account
- sets up a reusable function to grab the stripe status and update the cached data on the org
- adds additional hooks for updating subscription data


dheniges (changes requested):
Code looks excellent considering the complexity 👍 

The checkActiveOrgStatus middleware needs updating to use the new ORG_SUBSCRIPTION_STATUS constants though I think. If you register a new org you'll see the errors because it's checking against the STRIPE_STATUS

jjidt:
@dheniges updated, with the TRIALING_PENDING_ACTIVE status allowed too

dheniges (changes requested):
Seed data needs updating to work with new ORG_SUBSCRIPTION_STATUS values too

dheniges (approved):
:+1:

PR #301, opened 2019-11-27
Error logs and alerts

dheniges:
Requires all the other logging PRs.

- Logs caught errors
- Logs uncaught errors

Setup test alarm with email so we can get notified about the uncaught errors. Investigating an alert for heightened error rates


api/src/libs/middleware/apiErrorHandler.js
```diff
@@ -56,8 +58,8 @@ export default function(error, req, res, next) {
 
   const responseMessage = publicMessage || exposedMessage || getPublicMessage(responseCode)
 
-  //TODO handle conditional logging here for different environments
-  console.error(error.stack)
+  logger(LOG_LEVELS.WARNING, { error })
```
jjidt: Do we want log level to be conditional on responseCode, where 500s are logged on the ERROR level?

jjidt (commented):
One question

jjidt (approved):
🎉

PR #302, opened 2019-11-27
New Format and Content menus

jcontonio:
* Creates new DocumentMenuComponent so it's easy to spin up more menus in the document toolbar
* Adds formatting menu
* Changes ContentMenu to use the new DocumentMenuComponent
* Fixes the two text editor problems (inserting link, undo/redo)


dheniges (commented):
Screen Shot 2019-11-27 at 4 59 17 PM

This is awesome! Only issue I found is the height if the window is too small, I can't scroll up and see the upper options

Screen Shot 2019-11-27 at 5 04 56 PM

dheniges (approved):
Yay

PR #303, opened 2019-11-28
Adjusting field height on drag and drop

jcontonio:
* Just the functional bit, will add styling in another PR


dheniges (approved):
Works way better with this

PR #304, opened 2019-12-02
Styling code blocks

jcontonio:
Styles code blocks a little better


dheniges (commented):
Screen Shot 2019-12-03 at 2 52 14 PM

Definitely looks much nicer. Is it possible/relatively easy to fix the above scenario?

jcontonio:
> Screen Shot 2019-12-03 at 2 52 14 PM
> 
> Definitely looks much nicer. Is it possible/relatively easy to fix the above scenario?

@dheniges nope, not with this editor

dheniges (approved):

PR #305, opened 2019-12-02
Updating field settings UI

jcontonio:
* Makes image settings UI much nicer
* Fixes Safari button clipping issues in drag mode
* Breaks out field settings css into own file


dheniges (approved):
👍

PR #306, opened 2019-12-02
Title length and document mode cleanup

jcontonio:
* Adds ellipses overflow to long document names (make a really long document title, resize window)
* Takes the user out of document edit mode when switching routes


dheniges (approved):
:+1:

PR #307, opened 2019-12-02
default to fetching only published docs

jjidt:
- only return published docs by default
- return all docs for web


dheniges (approved):
:+1:

PR #309, opened 2019-12-03
Adds remove account UI

jcontonio:
* Adds Remove Account UI in Billing section
* Adds "confirmedText" as an option to the Confirm component forcing you to type something to continue
* Hits the cancel account endpoint


api/src/coordinators/billing.js
```diff
@@ -9,6 +9,7 @@ import { PLANS, MULTI_USER_DEFAULT_SEAT_COUNT, STRIPE_STATUS, ORG_SUBSCRIPTION_S
 import { getOrgSubscriptionStatusFromStripeCustomer } from '../libs/orgSubscription'
 
 const formatBilling = (customer, userCount) => {
+  console.log(customer.subscriptions.data[0])
```
jjidt: remove

web/src/client/js/components/Admin/AdminBilling/AdminBillingContainer.js
```diff
@@ -1,24 +1,55 @@
...
+const AdminBillingContainer = ({ deleteOrganization, uiConfirm }) => {
   const { data: currentOrg } = useDataService(dataMapper.organizations.current())
+  const { data: orgBilling } = useDataService(dataMapper.organizations.billing(), [currentOrg.plan])
+
+  function deleteAccountHander() {
+    const message = (
+      <>
+        <p className="danger">Delete your account entirely?</p>
+        <ul>
+          <li>You will be able to log in for the next X days to undo this account.</li>
```
jjidt: do you want to update these 'x' values? it will be conditional on whether they are TRIALING/TRIALING_PENDING_ACTIVE in which case their account will be immediately canceled, otherwise we can use the subscription.currentPeriodEnd value for 'x'

jjidt (changes requested):
Cancellation works, just a couple things.

Also, we need to fix some things with the flow, right now a 'TRIALING' user can cancel their account and they get 'INACTIVE' status, but this just results in a billing message, and they can enter billing info and get 'ACTIVE' status, which shouldn't be possible.  We need to completely lock out 'INACTIVE' users from the API and give them an 'Account removed, please contact blah blah if this was an error' message of some kind.

I think we can go ahead and merge this as soon as it's approved and do that in another PR, I think it'll involve passing in a set of acceptable statuses to certain API routes

web/src/client/js/components/Admin/AdminBilling/AdminBillingContainer.js
jcontonio: Oh good point, I'll add that in

jcontonio: @jjidt solved all of this, thanks

jcontonio:
> Cancellation works, just a couple things.
> 
> Also, we need to fix some things with the flow, right now a 'TRIALING' user can cancel their account and they get 'INACTIVE' status, but this just results in a billing message, and they can enter billing info and get 'ACTIVE' status, which shouldn't be possible. We need to completely lock out 'INACTIVE' users from the API and give them an 'Account removed, please contact blah blah if this was an error' message of some kind.
> 
> I think we can go ahead and merge this as soon as it's approved and do that in another PR, I think it'll involve passing in a set of acceptable statuses to certain API routes

I've broken out Notices and Cancel into their own components, and I am taking care of the case of a TRIALING user canceling, then adding payment. They can no longer do that (at least in the UI).

jjidt (approved):
👍 This looks good, I went ahead and pushed up a fix for the incorrect date thing, where it now forces a refetch of the billing info on cancel, and uses the currentPeriodEnd as a temporary stand-in

Still some things with the subscription flow that we need to touch up, let's go through that together sometime this week

jcontonio:
Sweet, thanks JJ! Yeah let’s do that

PR #310, opened 2019-12-03
stop executing passport callback in err condition

dheniges:
Without this change, an uncaught exception would be thrown because express was trying to send a second response from apiErrorHandler and the login route. Now this correctly stops executing the login route and passes the error along the middleware chain so only apiErrorHandler responds


jjidt (approved):
👍

PR #311, opened 2019-12-03
doc name version

jjidt:
PR https://github.com/BonkeyBong/portway/pull/307 needs merged in first

- version control for document names
- stores the doc name on DocumentVersion
- defaults to published docs if no draft=true querystring passed
- gets published name for documents/:docId query as well (now returns a 404 if doc is not published or unpublished instead of a doc with an empty fields array)


dheniges:
JJ this has a BusinessConflict™

dheniges (approved):
:+1: tested, and none of my old docs seemed to break either

PR #312, opened 2019-12-05
doc search

jjidt:
Only enabled for fetching draft docs from the projects/:id/documents route
Also, only currently searching text fields, but can be expanded with other models

### DIFF
--- a/api/src/db/__testSetup__/factories/field.js
+++ b/api/src/db/__testSetup__/factories/field.js
@@ -15,7 +15,7 @@ const getFieldData = function(override = {}) {
 
 const createMany = async function(numberOfFields, override = {} ) {
   // default to a number field
-  const type = override.type || FIELD_TYPES.NUMBER
+  const type = override.type || FIELD_TYPES.STRING
   const model = FIELD_TYPE_MODELS[type]
   let value = override.value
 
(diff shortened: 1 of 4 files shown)


dheniges (changes requested):
Functionality is great, fuzzy search gives me fuzzy warm feelings. (Maybe the 2nd cup of coffee was a mistake)

I'd add a few more test cases:
- no match on any docs
- matching a text field value

jjidt:
Now with more tests ™️

api/src/db/__testSetup__/factories/field.js
```diff
@@ -12,7 +13,25 @@ const getFieldData = function(override = {}) {
   return { ...defaultProps, ...override }
 }
 
-const createMany = async function(numberOfFields, override) {
+const createMany = async function(numberOfFields, override = {} ) {
+  // default to a number field
+  const type = override.type || FIELD_TYPES.NUMBER
```
dheniges: tests are passing, but any concerns with changing this behavior to defaulting to number vs string field?

jjidt: we can do string, I just noticed that more tests were using numbers, I can change it back and make sure everything still works

dheniges (approved):
🌮

PR #313, opened 2019-12-05
Fixing up creation of document field

jcontonio:
* Fixes bug when creating a new document with an empty field. It was invisible before and not focusing
* Updates createDocument function with some options so we can be more exact


dheniges (approved):
😸

PR #314, opened 2019-12-05
Free Account Setup

dheniges:
- New admin route for creating a free multi user plan org
- New stripe subscription plans for free tiers
- Updates web to not show payment info on free plans, and message about contacting us to upgrade
- Example payload in the controller
- Secured by secret admin key
- Can invite > default number of users (currently 5), but cannot later add more in UI
- Free accounts can invite up to 5 users if not added via endpoint
- Adds free single user plan but no way to create an org with that currently, can add later


api/src/coordinators/demoSignup.js
```diff
@@ -0,0 +1,83 @@
...
+
+  const orgUsers = await Promise.all(
+    users.map(async (user) => {
+      const { name, email } = user
+      const resetKey = passwordResetKey.generate()
+      const businessUser = await BusinessUser.create({
+        name,
+        email,
+        orgId: organization.id,
+        orgRoleId: user.role,
```
jjidt, line 41: seems like we could end up with multiple users with the OWNER role who aren't necessarily referenced by the ownerId on the org.  So might want to iterate over the passed in users and ensure there's only one owner

web/src/client/js/components/Admin/AdminBilling/AdminBillingComponent.js
```diff
@@ -35,11 +35,13 @@ const AdminBillingComponent = ({ organization }) => {
         </section>
         <hr />
       </OrgPlanPermission>
-      <section id="payment" className={sectionClasses}>
-        <h2>Payment information</h2>
-        <AdminPaymentContainer />
-      </section>
-      <hr />
+      <OrgPlanPermission acceptedPlans={[PLAN_TYPES.SINGLE_USER, PLAN_TYPES.MULTI_USER]}>
```
jjidt: This keeps us from upgrading our seed user bonkey@bong.com, do we give them a plan: SINGLE_USER in the seed data? not sure what effects that will have if they don't have the matching stripe subscription

jjidt (changes requested):
Works great, just a couple things

Also, wondering if we should put this at /admin/ instead of /api/admin/ then we could avoid the loader altogether and never accidentally mess with any admin routes? thoughts?

web/src/client/js/components/Admin/AdminBilling/AdminBillingComponent.js
dheniges: I've been thinking about our seed data, I'm not sure how useful it is and will be going forward due to the reliance on stripe. If we want to test a lot of projects/documents/fields we have these beautiful APIs that makes that pretty easy to generate. We can always test signing up for a new account and org locally with the existing app code, and maybe in the future we add another admin route that would auto-verify the seed user? For the very moment I guess I will allow `null` as an acceptedPlan

api/src/coordinators/demoSignup.js
dheniges: added validation for that now

web/src/client/js/components/Admin/AdminBilling/AdminBillingComponent.js
dheniges: added `null` support and tested with seed data

dheniges:
I moved the admin controller to `/admin` and outside the loader, agree it might be easy to do something we don't mean to do in there otherwise

api/src/coordinators/demoSignup.js
```diff
@@ -0,0 +1,89 @@
...
+  )
+
+  const customer = await stripeIntegrator.createCustomer({
+    name: organization.name,
+    description: `Customer for Org Owner: ${owner.email}`,
+    email: owner.email
+  })
+
+  await BusinessOrganization.updateById(organization.id, {
+    ownerId: owner.id,
```
jjidt: caught one other thing here, it looks like this is pulling the id from the passed in user object and not from the saved db user, so the ownerId isn't getting set

jjidt (changes requested):

jjidt (approved):
🎉 🌮

PR #315, opened 2019-12-05
Document options

jcontonio:
* Adds Unpublish and Delete to the Document UI itself
* Adds other responsive issues for Document UI
* Removes flexbox from editor because of weird issue

### DIFF
--- a/web/src/client/js/components/Button/Button.js
+++ b/web/src/client/js/components/Button/Button.js
@@ -1,0 +1,1 @@
+export { default as ButtonGroup } from './ButtonGroup'
--- a/web/src/client/js/components/Button/Button.js
+++ b/web/src/client/js/components/Button/Button.js
@@ -1,1 +1,5 @@
+// Eventually changing all buttons to use specific components so there is less
+// confusion or reliance on adhering to css classes
+// ButtonGroup is the first one, but there will be others
+
 export { default as ButtonGroup } from './ButtonGroup'
--- a/web/src/client/js/components/DocumentToolbar/DocumentToolbarComponent.js
+++ b/web/src/client/js/components/DocumentToolbar/DocumentToolbarComponent.js
@@ -57,7 +57,7 @@ const DocumentToolbarComponent = ({
           </>
           }
 
-          {documentMode === DOCUMENT_MODE.EDIT &&
+          {documentMode === DOCUMENT_MODE.EDIT && document.lastPublishedAt !== null &&
           <>
           Last published:&nbsp;
           <span title={moment(document.lastPublishedAt).format('MMMM do, YYYY - h:mma')}>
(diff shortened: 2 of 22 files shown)


web/src/client/js/components/Button/Button.js
```diff
@@ -0,0 +1 @@
+export { default as ButtonGroup } from './ButtonGroup'
```
dheniges, line 5: Is this file needed? If so, worth adding a comment

web/src/client/js/components/DocumentToolbar/DocumentToolbarComponent.js
```diff
@@ -38,15 +48,37 @@ const DocumentToolbarComponent = ({
...
+          <DropdownComponent
+            align="right"
+            autoCollapse={true}
+            button={{
+              className: 'btn--small btn--with-icon',
+              icon: <CaretIcon fill="#fff" />
+            }}>
+            <DropdownItem
+              className="btn--small"
+              label="Unpublish document"
```
dheniges: This button should get disabled if the doc hasn't been published, like it does in the document list menu

web/src/client/js/components/DocumentToolbar/DocumentToolbarComponent.js
```diff
@@ -38,15 +48,37 @@ const DocumentToolbarComponent = ({
...
-          {!isPublishing && <PublishIcon fill="#ffffff" />}
-          <span className="label">Publish {publishString}</span>
-        </button>
+          small
+          title={publishButtonTitle()}
+        >
+          <button className="btn btn--small btn--with-icon" onClick={publishDocumentHandler}>
+            {isPublishing && <SpinnerComponent width="12" height="12" color="#ffffff" />}
+            {!isPublishing && <PublishIcon fill="#ffffff" />}
+            <span className="label">Publish {publishString}</span>
```
dheniges: This publish button is no longer disabled after you publish a document and haven't made any changes

web/src/client/js/components/DocumentToolbar/DocumentToolbarContainer.js
```diff
@@ -41,21 +50,48 @@ const DocumentToolbarContainer = ({ isCreating, isPublishing, publishDocument })
...
+    const confirmedLabel = `Yes, delete this document`
+    const confirmedAction = () => { deleteDocument(projectId, document.id, history) }
+    uiConfirm({ message, confirmedAction, confirmedLabel })
+  }
+
+  function unpublishDocumentHandler() {
+    const message = (
+      <>
+        <p>Unpublish the document <span className="highlight">{document.name}</span>?</p>
+        <p>If you are using this document in a live application, it will be removed.</p>
```
dheniges: Probably worth moving this text/markup into a shared file so it's not repeated?

dheniges (changes requested):
A few feedback items on this.

What do you think UX-wise? The "Unpublish" being an arrow menu option above the "Publish" button makes sense to me, but having the "Delete Document" in that same menu isn't as intuitive to me. Took me a few moments to notice the menu next to the button when I booted up this PR.

web/src/client/js/components/Button/Button.js
jcontonio: Will add comment

dheniges (approved):
This feels nice and makes sense to me :+1:

PR #316, opened 2019-12-06
allow empty string for field value

jjidt:


jjidt:
```
"errorDetails": [
        {
            "message": "\"value\" must be a string",
            "key": "value"
        },
        {
            "message": "\"value\" must be a number",
            "key": "value"
        }
    ]
```

Error messages now look like this, since it's an array of allowed types
before it was passing the initial string check, but failing the empty string validation, and thats why we just had the number message

jcontonio:
I don't understand why "value must be a number" is coming back at all on a text field? We had both "string can't be empty" and "value must be a number" before. Are we still getting the value must be a number?

jcontonio (approved):
Can update text field with empty value 👍

jjidt:
That's the default error behavior when using Joi.alternatives, which is currently only allowing a string or number as a possible value. For an update, we don't know the type of field at this point, because that was set during creation and they could just be sending { value: 'something' }. The alternative would be to not validate the value initially at all and then do a custom validation by type somewhere later when we have already retrieved the field object.

PR #318, opened 2019-12-06
Text area tweaks

dheniges:
@jcontonio take a look, modified your #313 PR to move logic into a single option for creating a document with a body or not. See what you think


jcontonio:
Ah yeah, I like that 👍

jcontonio (approved):

PR #319, opened 2019-12-07
New Menus

jcontonio:
* Removes Dropdown components completely
* Adds new Popper, PopperGroup, Menu, MenuHeader, MenuItem components
* Makes all the things consistent
* Popper is a temporary name until I replace Popovers
* Fixes the height of the format menu on short screens

Also adds initial search actions, because I started there but then realized this had to be done


web/src/client/js/components/Admin/AdminPlanSelector/AdminPlanSelectorComponent.js
```diff
@@ -54,10 +55,12 @@ const AdminPlanSelectorComponent = ({
         Your plan: <span className="admin-plans-selector__title">{planTitle}</span>
       </h2>
       {hasFreePlan ?
-        <p>You are currently on a free plan. If you would like to upgrade, please contact us.</p>
+        <p>
+          You are currently on a free plan. If you would like to upgrade, please
+          <a href={SUPPORT_EMAIL}>contact us</a>.
```
dheniges: should have `mailto:`?

dheniges (changes requested):
Everything seems in order here except none of the popovers close when you click on the element that opens them. So if you open the content menu, you should be able to close it by clicking on the "+" icon a second time (this was the previous behavior)

web/src/client/js/components/ContentMenu/ContentMenuComponent.js
```diff
@@ -35,7 +35,7 @@ const ContentMenuComponent = ({ createFieldHandler }) => {
         aria-controls="content-menu"
         aria-label="Add a field"
         className="btn btn--blank btn--with-circular-icon document-menu__button"
-        onClick={() => setExpanded(true)}
+        onClick={() => setExpanded(!expanded)}
         onFocus={() => setExpanded(true)}
```
dheniges: I think the onClick and onFocus may be interfering with each other? Menu will open and close rapidly on the first click, then second click it opens

dheniges (approved):
:+1:

PR #323, opened 2019-12-10
🕵🏻‍♀️ Search results!

jcontonio:
* Hooks in search actions / reducer, focused on documents now
* Adds up/down navigation of search results


dheniges (commented):
Yay search! But...

I think this UX is a bit confusing. It's not clear that documents are only searched in the current project you're looking at. I think having the search work like a document filter in the document list bar would be much more intuitive.

And as you were saying, this feels like a slight variation on the existing project selector. My inclination at this point is to keep the project selector that already has search, and add the document search bar. Or update this to replace the project selector

What do you think?

Other things:
- you can't close the search menu without clicking a result in it or navigating to a different url
- the search string is preserved, but the search results are not. Enter a search, navigate somewhere else, click the search bar. It will show the correct project results but not any document results
- Project title search is case sensitive
- If you type a word and hit enter in the search bar it refreshes the page and you lose your search string

web/src/client/js/components/Document/DocumentContainer.js
```diff
@@ -29,18 +29,26 @@ const DocumentContainer = ({
...
+  /**
+   * If we're creating a document, render nothing
+   */
+  if (isCreating || match.params.documentId === PATH_DOCUMENT_NEW_PARAM) {
+    return null
+  }
+
   // if we don't have a documentId, we shouldn't be loading this component
-  if (match.params.documentId == null) {
+  if (match.params.documentId === null) {
```
dheniges: `==` catches `undefined` too in this case, as `undefined == null` is true but `undefined === null` is false

web/src/client/js/components/Document/DocumentContainer.js
```diff
@@ -29,18 +29,26 @@ const DocumentContainer = ({
...
     return null
   }
 
   //if the document id isn't a valid number, redirect to 404
-  if (isNaN(match.params.documentId)) {
+  if (!isCreating && match.params.documentId && isNaN(match.params.documentId)) {
```
dheniges: this `!isCreating` check is superfluous as you've already returned null when isCreating is true, and the `match.params.documentId` check is also already handled above

web/src/client/js/components/Document/DocumentContainer.js
```diff
@@ -29,18 +29,26 @@ const DocumentContainer = ({
...
     return <NoDocument />
   }
 
   //if we're done loading things but the data never arrives, assume 404
-  if (documentLoading === false && !document) {
+  if (!isCreating && documentLoading === false && !document) {
```
dheniges: same - isCreating check is already handled

web/src/client/js/components/Document/DocumentContainer.js
```diff
@@ -50,19 +58,13 @@ const DocumentContainer = ({
...
-    return null
-  }
 
   /**
    * If there is no document and we are not creating: true, then we render
    * a helpful message
    */
 
-  if (typeof match.params.documentId === 'undefined' || match.params.documentId === PATH_DOCUMENT_NEW_PARAM) {
+  if (!isCreating && typeof match.params.documentId === 'undefined' || match.params.documentId === PATH_DOCUMENT_NEW_PARAM) {
```
dheniges: I don't think it's possible to reach this condition

web/src/client/js/reducers/search.js
```diff
@@ -0,0 +1,52 @@
...
+  isSearching: false,
+  searchTerm: null,
+  searchResultsByDocumentId: null,
+}
+
+// Search results
+export const search = (state = initialState, action) => {
+  switch (action.type) {
+    case ActionTypes.SEARCH_CLEAR: {
+      // Reset on route change
```
dheniges, line 13: This seems like it should reset on route change in some circumstances (switch projects and the search results from the last project show up)

dheniges (changes requested):
The 404 stuff needs a little bit more love, I can grab that in a separate PR if you'd rather.

Search works, but it carries results over between projects because it doesn't get cleared. It seems like search state (and results) should clear as soon as you navigate away from a project, but not sure how hard that is to implement. Right now you can search, then switch projects, and the search + results from the old project remain.

Also I dunno about the search requiring a click + covering up the new document. It's kinda cute, but seems too easy to get into a "how the hell do I add a new document!?!?" state. See screenshot below. All my favorite apps have an open search bar I can start typing into immediately, with the exception of some mobile apps.

Other things:
- spacing is left in the header for the first search bar, makes the "upgrade account" float oddly
- bullets got added to the content menu list (see screenshot)

Screen Shot 2019-12-16 at 4 35 08 PM

Screen Shot 2019-12-16 at 4 30 46 PM

dheniges:
Updated the 404 stuff. Everything as of this comment looks good except the double search clear function and overlap of the search bar into the search clear button.

dheniges (approved):
Quick merge it fast before anyone notices

PR #324, opened 2019-12-10
Updates for taking prod live

dheniges:
- New annotations and configuration for integrating Lets Encrypt with the cert-manager
- cert-manager is great but it's got an alpha API and will continue to change rapidly. I think it's worth the pain, as it will keep working in old versions unless Lets Encrypt changes their API, which seems unlikely given how many sites rely on it
- The dev cert-manager setup is likely broken now. This can be reviewed and merged and I will work on fixing dev before the next deployment
- Adds letsencrypt issuer k8s resources for cert-manager to repo


jjidt (approved):
👍

PR #326, opened 2019-12-10
Disabling sign-up temporarily

jcontonio:
Hides sign-up stuff all over the Express views so we don't get any random sign-ups until we're ready


dheniges (approved):
:+1:

PR #327, opened 2019-12-10
Project/Doc 404 handling on web

jjidt:


jcontonio:
I tried the following and got the 404 page:
http://localhost:3000/d/project/110000/document/1200019191

I tried the following and didn't get the 404 page, but got the notification
http://localhost:3000/d/project/12

Renders the project, but oddly, and no 404
http://localhost:3000/d/project/undefined
http://localhost:3000/d/project/null

I think we should set up a route at the top level for the 404 page and redirect all of these cases there, instead of rendering the project view. Is this what you were talking about this morning? That a redirect is a problem?

jjidt:
I'll give it a try with a top level 404 and see if the redirect still has issues

jjidt:
@jcontonio was still having infinite loop issues, but found that returning a router Redirect component takes care of it, check out the behavior and see if it's along the lines you were thinking

jcontonio:
Only thing I am seeing is http://localhost:3000/d/project/110000/document/asd doesn't return a 404 anymore, and no errors - the rest of the example URLs work, just not an invalid document id

jjidt:
@jcontonio ready for more of that sweet sweet QA

jcontonio:
Ok, after we chatted through this:

1. /project/blah -> render `/components/ProjectList/NoProject.js`
2. /project/110000/document/blah -> render `/components/Document/NoDocument.js`
3. anything else -> renders catch all 404 at top route (I pushed this up)

hopefully without any redirects or notifications

jjidt:
ok, I think this is what we want now, ready for another look

jjidt:
added the redirect from root to /projects to this PR to avoid conflicts

dheniges (commented):
Scenarios I tested all work. @jcontonio do you want to style the 404s here or backlog that?

dheniges (approved):
👏

PR #332, opened 2019-12-11
cover cases where org or user aren't in state

jjidt:
I couldn't reproduce the issue from the deployed app locally, but it seems like we could have a race condition of some kind so I added checks that the items we're looking at are populated

Make sure the web org permissions access stuff is working, then after merging we'll get this up on dev


dheniges:
@jjidt not quite sure from the PR description - is this ready for review?

dheniges (approved):
:+1:

PR #336, opened 2019-12-16
Deploy 0.9.4 to dev and prod

dheniges:
Thanks to the codefresh bug, this will trigger 4 builds!

Fortunately the kubes will handle it like a champ.


jjidt (approved):
👍

PR #337, opened 2019-12-16
Redirect back to url after login

dheniges:
- If you go to a saved url in the app but you're not logged in, this will redirect you back to that url after logging in
- Removes unused alias middleware on the web server


jcontonio (approved):
Nice!

PR #339, opened 2019-12-17
Logged in user in web server

dheniges:
- Creates a `req._portway` namespace for web server requests
- Populates `req._portway.loggedIn` as `true` if the user has a valid session
- Redirects from /sign-in to app projects page if user logged in

Requires #337


jjidt (approved):
👍

PR #341, opened 2019-12-18
Admin seats management

dheniges:
- Fixes bug that redirected admins to org management page, then back to projects
- Adds org seats endpoint to fetch org seat info. Added resource type and role permissions so admins and owners can access it
- Replaced billing subscription fetching in UI with seats endpoint


api/src/controllers/orgSeats.js
```diff
@@ -16,25 +15,37 @@ const paramSchema = Joi.compile({
...
-  const org = await BusinessOrganization.findSanitizedById(orgId)
+  return perms((req) => {
+    return {
+      resourceType: RESOURCE_TYPES.SEAT,
+      action: ACTIONS.READ
+    }
+  })(req, res, next)
+}
+
+const conditionalUpdatePerm = async (req, res, next) => {
```
jjidt: can we un-asyncify this one too?

web/src/client/js/reducers/organizations.js
```diff
@@ -130,6 +132,30 @@ export const organizations = (state = initialState, action) => {
       }
     }
 
+    case ActionTypes.REQUEST_ORGANIZATION_SEATS: {
```
jjidt, line 140: we could update the seatsById and loading: seatsById when the entire billing object comes back too, what do you think?

dheniges: good idea, added

jjidt (approved):

PR #342, opened 2019-12-18
Adding user search

jcontonio:
* Adds user search back in
* Adds forwardRef to FormField so we can refer to it and clear it manually
* Removes default `type="search"` styling from input fields


web/src/client/js/libs/dataMapper.js
```diff
@@ -206,15 +216,18 @@ export default {
...
         fetchAction: searchByName(partialNameString),
         getLoadingStatusFromState: (state) => {
           return state.users.loading.byNameString[partialNameString]
         },
         getDataFromState: (state) => {
           const ids = state.users.userSearchResultIdsByNameString[partialNameString]
-          const users = ids && ids.map(id => state.users.usersById[id])
-          return users
+          const userSearchResults = ids && ids.map(id => state.users.usersById[id])
+          // @todo JJ make this work! =)
```
dheniges, line 228: Is this todo to add pagination support? Should go in asana

dheniges (commented):
Screen Shot 2019-12-18 at 10 49 57 AM
^ highlight colors are weird on the search field, for comparison below is the org name field:

Screen Shot 2019-12-18 at 10 50 13 AM

If you start typing you get the bold highlight color, and then if you click on the field it changes to the darker blue, narrower highlight. I assume that is not on purpose?

web/src/client/js/libs/dataMapper.js
jcontonio: Made ticket for 1.0 and assigned to @jjidt because he's the only one who knows what's going on there

dheniges (approved):
:+1:

PR #343, opened 2019-12-18
remove 'No Document' flash when switching between projects and docs

jjidt:
- the currentResource function was returning a loading state of false rather than null for non-request situations, resulting in the components thinking the fetching had been completed
- the url changes before the useDataService code runs, so the component will rerender with the previous project's document before getting the right one, adds an additional check to make sure the project's id matches the url param


dheniges (commented):
the currentResource change appears good. There was still a janky flash when switching between docs, whether or not the doc had been loaded. See my PR into this PR

dheniges (approved):
the combination of these two PRs has created perfection 💯

PR #344, opened 2019-12-18
Render blank doc when document loading/switching

dheniges:
This prevents an unsightly flash when switching between docs, where the bottom document toolbar renders first, then the document title and text area


jjidt (approved):
looks way better, nice idea 👍

PR #346, opened 2019-12-18
Moving Back to on Users view

jcontonio:
Moves "Back to Users" from above the navigation to above the content, consistent with other Panel UIs. This is on the User view in Admin Users


dheniges (changes requested):
It works, but the arrow is practically invisible:

Screen Shot 2019-12-18 at 11 28 51 AM

dheniges (approved):
Nevermind it's just me. Looks good

PR #347, opened 2019-12-18
api versioning

jjidt:
- v1 versioning for all current /api routes
- manually changed each web action to point to v1 url rather than adding a global prefix, so we won't have to go back through and do that when the first v2 route rolls out
- need to update stripe hooks to point the v1/ when this is deployed


dheniges (changes requested):
Everything I tested works but setting initial password in the signup controller isn't prefaced with v1 (`web/src/server/controllers/sign-up`)

jjidt:
nice catch, updated

dheniges (approved):
🎉

PR #349, opened 2019-12-18
Swapping button labels

jcontonio:
Makes sure the button labels are correct on Admin Org settings (wtf)


dheniges (approved):
accurately labeled buttons :+1:

PR #351, opened 2019-12-18
1Password Generator fix

jcontonio:
Removing debounce on pw change handler fixes the problem

Before:
Generate a new password and fill it here: http://localhost:3000/d/settings/security
The password's would both match, but react state was off, so it would tell you they didn't

After:
You can generate a password, fill it, and you won't get the warning


dheniges (approved):
:+1:

PR #352, opened 2019-12-19
Feature: Email templates

jcontonio:
Here is the first commit.

* We need mjml to generate the HTML for these
* We can either generate them outside of our app, and put .ejs files in here, or we can use the mjml node library to generate the HTML, such as:

```
import mjml2html from 'mjml'
/*
  Compile an mjml string
*/
const htmlOutput = mjml2html(`
  <mjml>
    <mj-body>
      <mj-section>
        <mj-column>
          <mj-text>
            Hello World!
          </mj-text>
        </mj-column>
      </mj-section>
    </mj-body>
  </mjml>
`, options)
```


dheniges (approved):
:jjidt:

PR #356, opened 2019-12-19
Fixing empty search results

jcontonio:
* Does a check for isSearching because if documents is empty, it was thinking it was a new, empty  project
* Sets searchResults var to null by default, and a check if it's valid vs an empty array


dheniges (approved):
Works great

PR #357, opened 2019-12-19
stripe doc update

jjidt:


dheniges (approved):
that's why I didn't setup the stripe hook with all the right events :)

PR #360, opened 2019-12-19
Removing TRIALING_PENDING_ACTIVE from trialing status

jcontonio:
Removes TRIALING_PENDING_ACTIVE from trialing statuses, because it should behave as ACTIVE.

This is when a user is in TRIALING, and adds payment information. They should become "active" in our UI.


dheniges (approved):
:+1: saw this issue before, new behavior is as expected

PR #361, opened 2019-12-19
On cancel, log user out once

jcontonio:
* When canceling an account, log the user out 
* Don't log the user out every time they go to the billing page


dheniges:
Definitely an improvement on not forcing logouts, but do we actually want to log out the user when they cancel their account? I'm leaning toward leaving them on the billing page so they can see the "Account canceled"

jcontonio:
Sounds good to me!

dheniges (approved):
:+1:

PR #362, opened 2019-12-19
Adding trial end language

jcontonio:
Adds language for trial ending

### DIFF
--- a/web/src/client/js/components/Admin/AdminNotices/AdminNoticesComponent.js
+++ b/web/src/client/js/components/Admin/AdminNotices/AdminNoticesComponent.js
@@ -32,7 +32,7 @@ const AdminNoticesComponent = ({ logoutAction, organization, subscription }) => 
     <div className="admin-notices">
       {TRIALING_STATUSES.includes(organization.subscriptionStatus) &&
       <div className="admin-notices__notice">
-        <h2 className="admin-notices__notice-title"><InfoIcon width="22" height="22" /> Trial period</h2>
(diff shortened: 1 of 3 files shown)


web/src/client/js/components/Admin/AdminNotices/AdminNoticesComponent.js
```diff
@@ -27,12 +27,16 @@ const AdminNoticesComponent = ({ logoutAction, organization, subscription }) =>
...
+  const trialEnds = subscription.trialEnd
+  const trialEndsInDays = moment.unix(trialEnds).diff(moment(), 'days')
 
   return (
     <div className="admin-notices">
       {TRIALING_STATUSES.includes(organization.subscriptionStatus) &&
       <div className="admin-notices__notice">
-        <h2 className="admin-notices__notice-title"><InfoIcon width="22" height="22" /> Trial period</h2>
+        <h2 className="admin-notices__notice-title">
+          <InfoIcon width="22" height="22" /> Trial {trialEnds && <>ends in {trialEndsInDays} days, {moment(trialEnds).format('MMMM Do')}</>}
```
dheniges: I would add "trial ends in X days, **on** [date]"

(but don't bold the on, just adding "on" to that sentence)

dheniges (changes requested):
The date calculation isn't correct on this one:
Screen Shot 2019-12-19 at 4 06 00 PM

jcontonio:
I was using the wrong Moment method, it will now do "in 10 days" or "in a few seconds", etc, using the fromNow() method.

https://momentjs.com/docs/#/displaying/fromnow/

dheniges (approved):
:+1:

PR #366, opened 2019-12-20
update billing flow diagram

jjidt:


dheniges (approved):
Still contains the black abyss 👍

PR #367, opened 2019-12-20
Removing google font

jcontonio:
* We don't use serifs anywhere in the app anymore, so removing from head / type files
* Commented them out so I can bring them over to the marketing site easily, will remove comments then


dheniges (approved):
:+1: app still renders things

PR #368, opened 2019-12-20
Fixing disabled form check

jcontonio:
* Adds useEffect to make sure the Form re-renders when disabled property changes


dheniges (approved):
Password change form works now on the first try

PR #369, opened 2019-12-20
Fixing new document name

jcontonio:
This shouldn't have had a border:

Screen Shot 2019-12-20 at 12 50 34 PM


dheniges (approved):
Mmm yes, much nicer

PR #371, opened 2019-12-20
Replacing most colors with proper variables

jcontonio:
* Goes through and replaces most manual hex colors with css vars


dheniges (approved):
Things still have colors

PR #372, opened 2019-12-20
Adds lazy prop to images

jcontonio:
Does what it says. Big PR


dheniges (changes requested):
Screen Shot 2019-12-20 at 4 36 08 PM

jcontonio:
Updated, didn't see that warning

dheniges (approved):
Works now

PR #373, opened 2019-12-20
disable signup flag

dheniges:
Now signup can be disabled on prod, but enabled on dev and enabled locally! Woo!


jjidt (approved):
👍works great, I like the FLAG_ prefix for these env variables

PR #374, opened 2019-12-20
logging and stripehook awaiting

jjidt:
- Adds a logger for stripe events
- Actually await in controller so we know if any event-related updates were successful


dheniges (approved):
Good catch on the await

PR #382, opened 2020-01-03
Login & Sign-up screens

jcontonio:
* Removed any trace of Bulma and the old thinking of having a site in here
* Made log-in, sign-up, and other screens look nicer 

Screen Shot 2020-01-02 at 5 15 54 PM


web/src/server/views/user/sign-up.ejs
```diff
@@ -0,0 +1,17 @@
+<%- include('../partials/head', { title: title, permalink: permalink, robots: 'index, nofollow' }); %>
+<main>
+  <div class="container">
+    <img src="/images/logo.svg" alt="Portway Logo" width="85" />
+    <h2>Create your free account</h2>
+    <p>Start your <b>free, 30-day trial</b> now. <b>No credit card required.</b></p>
```
dheniges, line 6: The trial is currently 10 days in the API stripe config. We can change it to whatever, but should keep it consistent.

web/src/server/views/partials/footer.ejs
```diff
@@ -0,0 +1,8 @@
+<footer>
+  <ul>
+    <li><a href="https://getportway.com/">Portway</a></li>
+    <li>Privacy</li>
+    <li>Terms of Service</li>
+    <li><a href="https://getportway.com/support">Support</a></li>
```
dheniges: should use the constant so the support url, whatever that will be, only needs to be updated in one spot

dheniges (changes requested):
The "complete registration" (aka pick a password) after a new signup is broken. No requests are made, it's just a spinner

web/src/server/views/partials/footer.ejs
jcontonio: What's the best way to do this? Can I import the constant at the Express level and make it available globally somehow?

dheniges (approved):
:+1:

PR #383, opened 2020-01-03
fix country value

jjidt:
This gets the address info set correctly on the card, which comes back on the result object. We're still not storing any address info on customer billing, not sure if we'll want to do that.


dheniges (approved):
:+1:

PR #385, opened 2020-01-03
fix image move and copy

jjidt:


dheniges (approved):
:+1: drags images between docs like a champion

PR #386, opened 2020-01-04
stripe org name update

jjidt:


dheniges (changes requested):
I'm getting an error in the client, when the organization is updated the API is returning an empty 200 response but the client is expecting the org object

dheniges:
The org name does update correctly in Stripe though

jjidt:
updated

dheniges (approved):
🎉

PR #388, opened 2020-01-06
update user metadata on source create or update

jjidt:
the customer address object is required to be complete, so storing this as metadata, which is also searchable in the UI, tested with update and create


dheniges (approved):
:+1:

PR #391, opened 2020-01-07
Adding project deletion to project settings

jcontonio:
* Adds Delete Project to Project Settings
* Makes user confirm with project name

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -4402,18 +4402,18 @@
       "dev": true
     },
     "autoprefixer": {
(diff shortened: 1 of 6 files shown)


dheniges (changes requested):
This works, but I feel like the delete experience should be identical whether you do it from the list page or the project settings. Currently the text and verification processes are different

jcontonio:
@dheniges updated

dheniges (approved):
:tada:

PR #392, opened 2020-01-07
making catchall client error more generic

dheniges:
Browsers seem to log 404 as the response code from a CORS issue, and it sort of fits, the resource may exist but the client can't find it 🤷‍♀ 

To test, in the API's app.js comment out the CORS section toward the top


jjidt (approved):
👍

PR #393, opened 2020-01-07
Adding success message for re-invite

jcontonio:
Adds a spinner, then success icon when re-inviting a user


dheniges:
This works, but if you have 10 users it shows 10 spinners next to each user and then 10 green check marks on success. I'm guessing that's not quite what you're after here?

jcontonio:
@dheniges fixed

dheniges (approved):
Perfect

PR #395, opened 2020-01-07
fixing seat discrepencies on web

dheniges:
Basically, updating state.organization.seatsById to remove it when users are added or removed, and when org seats are changed on the plan.

Fixes all the issues I could find on web. Opening a new ticket for the fact you can create more users on the API than your plan allows


jjidt (changes requested):
When I create a user and navigate back to the billing page the current seats aren't updated

jcontonio:
Just went through, and yeah, that ^

Fixes the scenario I ran into earlier 👍

dheniges:
The problem with the billing page is the user counts aren't updated on the org subscription when users are added or removed. That has to get fixed on the API, which is a separate ticket

dheniges:
Ignore my previous comment. Forces refresh of org billing on user add/delete now

jjidt (approved):

PR #396, opened 2020-01-07
Redirecting to app if logged in

jcontonio:
Needed to add this to the new homepage

http://localhost:3000 should redirect if you're logged in


dheniges (approved):
Redirects as expected

PR #398, opened 2020-01-08
Puppeteer pipeline

dheniges:
It works! https://g.codefresh.io/build/5e1607fb21256f69d4d2d693?step=puppeteer_tests&tab=output

Need to setup the triggers after this gets merged, but it successfully runs in manual mode on Codefresh against dev


jjidt (approved):
👍

PR #403, opened 2020-01-08
reset password

jjidt:


dheniges (changes requested):
I'm getting a compile error in web:
```
ERROR in multi ./src/client/css/index.scss ./src/client/js/passwordReset.js webpack-hot-middleware/client?path=/__webpack_hmr&timeout=20000&name=passwordReset
portway-web | Module not found: Error: Can't resolve './src/client/js/passwordReset.js' in '/web'
```

web/src/server/views/user/password-reset.ejs
```diff
@@ -3,7 +3,22 @@
...
+          <label class="label is-small" for="email">Email</label>
+          <div class="field__control">
+            <input id="email" class="input" type="email" name="email" placeholder="email@domain.com"
+              autocomplete="section-login username">
+          </div>
+        </div>
+      </div>
+      <div class="field-container field-container--with-space">
+        <div class="field field--centered">
+          <input type="submit" class="button" value="reset your password">
```
dheniges: Should capitalize "Reset"

dheniges (changes requested):
It works, but two things to either update or backlog:
- If an email address doesn't exist, should tell the user when they submit the reset form
- I don't think we need a new confirmation page, a notification back on the home/login page should be good

jjidt:
Not sure about the first one, is that a security vulnerability to allow other users to see which email addresses are registered?  Updated the second with a flash message and fixed the button

dheniges (changes requested):
Thanks for the updates. I think you're right about the security issue, we probably shouldn't reveal whether the account exists. Still seeing:
- No flash message. Not sure if it's not plumbed on that login page, but the message doesn't show up
- Should catch the error if an email is entered without a corresponding user. We want to try to keep the uncaught errors solely to cases we're not expecting. I think a 204 response is fine in this case, so there's no difference between an existing or non-existing user.

jjidt:
@dheniges check out the logging I added and see if you think that's a good way to structure it, I'm still seeing the flash message, so we might have to screenshare to see what's going on for you

dheniges (approved):
:+1: I SEE GÜD

PR #405, opened 2020-01-09
Adding big invisible button

jcontonio:
* Adds Big Invisible Button™ to the end of the document, if the last field of a document is not a text field
* This should give the appearance of one big continuing document body
* Fixes a bug in naming new fields with the appropriate number, so now fields should get autonamed correctly in order (field-number-1, field-number-2, etc)


dheniges (changes requested):
This is purdy nice. Only issue I found is if you're in the edit mode to delete/reorder fields, the invisible button is still present and will create a new content field. I think the button should be disabled when in document edit mode

jcontonio:
Updated

dheniges (approved):
V. nice!

PR #406, opened 2020-01-09
Fixing roles buttons; Fixing focus styles

jcontonio:
* Updates OrgRolesDropdown and ProjectRolesDropdown to use the nice keyboard navigation stuff added in the Search PR
* Changes "Add user" form to use OrgRolesDropdown instead of React Select
* Updates some styling so that focus is different depending on mouse-mode or keyboard-mode, so that Menu's focus the same as inputs in forms

To test:

* Click on the "Full name" field in "Add user" form. You should be able to tab through the "Add user" form, and use the arrow keys to navigate the dropdown - focus styles should be consistent
* If you tab TO the "Add user" form (keyboard-mode), the focus should be consistent here too
* Teams under Project settings should all work the same too

### DIFF
--- a/web/src/client/css/inc/_buttons.scss
+++ b/web/src/client/css/inc/_buttons.scss
@@ -137,6 +137,36 @@ button,
     }
   }
 
+  &--form {
+    background-color: var(--button-white-background);
+    border: thin solid var(--button-white-border);
+    color: inherit;
+    &:not([disabled]) {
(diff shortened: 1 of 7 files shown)


dheniges (approved):
Well I can't test the org role dropdown on the teams page due to another bug, but this works nicely when adding a user 👍

PR #407, opened 2020-01-09
First email mjml template

dheniges:
- Setups up an mjml compilation step (`npm run compile-email`)
- async EJS render functions compiled on API startup
- Invitation email plugged in

Remaining feature items to be done:
- Need to add docs
- Need to do the rest of the emails


jcontonio (approved):
Seems good to me

PR #408, opened 2020-01-10
DND bug fix - Switching to useRef

jcontonio:
Fixes drag and drop bug of dragEnd events triggering immediately after dragStart. This is what was getting "stuck".

React useState was causing a re-render, which would put the DOM in a weird state on dragStart. Switching to useRef to keep track of the draggingElement seems to have fixed it.

### DIFF
--- a/web/src/client/js/components/Document/_Document.scss
+++ b/web/src/client/js/components/Document/_Document.scss
@@ -87,6 +87,7 @@
   }
 
   &__fields {
+    padding-bottom: var(--brand-width);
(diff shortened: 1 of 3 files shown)


dheniges (changes requested):
This applies the dragging styling to all the fields to make them short even when you're not dragging or not in the edit mode on a document. So long fields get cutoff:

Screen Shot 2020-01-10 at 2 08 39 PM

dheniges (approved):
Everything looks good here now 👍

PR #409, opened 2020-01-10
All emails templated

dheniges:
- mjml for all the emails. Emails need cleanup and styling, but I tested functionality on everything except payment succeeded and payment failed since those are stripe webhooks
- requires #407


jcontonio (approved):
I gets za emailz

PR #410, opened 2020-01-10
org data db deletion

jjidt:
to test, you'll need to add the DELETE organization perm for admins, the route is currently locked down and no one has that perm, we'll remove the endpoint after testing and when the job is up and running


dheniges (approved):
It deletes the org and associated things 👍

PR #411, opened 2020-01-10
Project teams user menu fix

jcontonio:
Fixes the user search dropdown on Project teams' "add user" section


dheniges (approved):
Fixes it 👍

PR #412, opened 2020-01-11
Doing single field focus

jcontonio:
If we are in a new document, or a document with one blank text field, clicking anywhere within the document should focus that field

### DIFF
--- a/web/src/client/js/components/DocumentFields/DocumentFieldsContainer.js
+++ b/web/src/client/js/components/DocumentFields/DocumentFieldsContainer.js
@@ -47,6 +47,25 @@ const DocumentFieldsContainer = ({
     setOrderedFields(fieldMap)
   }, [fields])
 
+  useEffect(() => {
+    // If we are in a new document, or a document with one blank text field,
+    // clicking anywhere within the document should focus that field
+    const fieldValues = Object.values(fields)
+    function documentClickHandler(e) {
+      // If we're clicking the document, focus the first text field
+      if (e.target.classList.contains('document')) {
+        const cm = document.querySelector('.CodeMirror').CodeMirror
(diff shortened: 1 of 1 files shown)


web/src/client/js/components/DocumentFields/DocumentFieldsContainer.js
```diff
@@ -47,6 +47,25 @@ const DocumentFieldsContainer = ({
...
+        cm.focus()
+      }
+    }
+    if (fieldValues.length === 1 && fieldValues[0].type === FIELD_TYPES.TEXT) {
+      document.addEventListener('click', documentClickHandler, false)
+      return function cleanup() {
+        document.removeEventListener('click', documentClickHandler, false)
+      }
+    }
+  }, [fields])
```
dheniges: I don't think you ever want to use objects as dependencies, because it won't actually compare object properties and values to see if it's changed. I think outside of the `useEffect` you'd want to do something like:
```
const fieldValues = Object.values(fields)
const hasOnlyOneTextField = fieldValues.length === 1 && fieldValues[0].type === FIELD_TYPES.TEXT
useEffect(() => {...}, [hasOnlyOneTextField])
```

dheniges (changes requested):
see comment on useEffect dependency

web/src/client/js/components/DocumentFields/DocumentFieldsContainer.js
jcontonio: Ah nice, thanks. I forgot about that. 
Right above this, we're doing another useEffect which is also looking at fields, but not sure what else we could compare there for a dependency?

dheniges (approved):
Beautiful

PR #413, opened 2020-01-13
fixing hook dependency on object in document fields container

dheniges:


jcontonio:
I get an error

```
 The final argument passed to useEffect changed size between renders. The order and size of this array must remain constant.

Previous: [1, 2, 3, 4, 5, 6, 7, 8, 9, 11, 130001, 130002, 130003]
Incoming: []
```

dheniges:
Well no variable length dependency arrays. I didn't realize there's already another callback handling setting order on the drag events, so this effect only has to fire when fields first load, or when a field is added or deleted, which using the fieldIds length will catch

jcontonio (approved):

PR #416, opened 2020-01-13
soft deleted user restore

jjidt:
we might have issues undoing the migration I added.  If there are items with null values in the database it will throw an error when trying to revert the allowNull constraint.


dheniges (commented):
This works, I can add a user to the same org or a new org. Question on the migration: seems like it shouldn't be reversible, so should the down() just do nothing?

jjidt:
yeah, that's what I was wondering, makes sense to me

jjidt:
updated with an empty promise

dheniges (approved):
v. nice

PR #417, opened 2020-01-14
send email on cancellation

jjidt:


dheniges (changes requested):
This is going to need to go into the feature email branch with an mjml template etc. via the email coordinator. I have a PR open with docs on that process

jjidt:
Screen Shot 2020-01-17 at 5 31 05 PM
Added an email template, not sure about the image assets, but tested with ngrok and the flow is working correctly

dheniges (approved):
:+1:

PR #418, opened 2020-01-14
support form zendesk endpoint on web

dheniges:
Uses zendesk API to create a ticket, and corresponding comment from the user with whatever data they put in the support form


dheniges:
Need to add the env var to the env var documentation too, oops

jcontonio:
~Need key in 1Password file, can't copy it from Zendesk's site~ Added a new one

jcontonio (approved):

PR #422, opened 2020-01-15
Read-only document mode

jcontonio:
* Removes all editing UI for documents if user is READER or if its a public project
* Fixes fields dependency bug in DocumentFieldsContainer


dheniges (approved):
I think this looks good.

I only noticed two things in readonly mode that don't really matter that much:
- If you click on a label for a string or number field, it highlights blue
- If you click the document title it turns black

PR #423, opened 2020-01-16
Initial asset storage tracking

dheniges:
- Creates generic resource tracking table for orgs
- Adds an asset type for tracking
- Increments value when assets added via add/update fields

The feature still needs:
- ability to remove assets and decrement tracker (requires changing publishing to copy asset, not just asset url)
- errors in the client to show up for image fields


api/src/db/models/organizationresourceusage.js
```diff
@@ -0,0 +1,17 @@
+module.exports = (sequelize, DataTypes) => {
+  const OrganizationResourceUsage = sequelize.define(
+    'OrganizationResourceUsage',
+    {
+      orgId: DataTypes.INTEGER,
+      resourceType: DataTypes.INTEGER,
+      value: DataTypes.TEXT
```
jjidt: this should be INTEGER right? not sure how the tests are passing validation

jjidt (changes requested):
Functionally works great! one request:

api/src/db/models/organizationresourceusage.js
dheniges: updated

jjidt (approved):

PR #424, opened 2020-01-17
React upgrade

jcontonio:
* Upgrades React, Linting, etc


dheniges (approved):
Clicked around quite a bit, didn't find anything broken with the upgrade. Also now I feel slightly more sane knowing the useEffect function was written correctly

PR #425, opened 2020-01-17
new api docs

jjidt:
This isn't 100% complete, added TODOs for query params.  We could probably split up controllers into separate files like I did with the models, but I think it gets us most of the way there


dheniges (approved):
Thanks for all the work on this JJ, I'm sure it wasn't the most exciting thing. 👍

PR #426, opened 2020-01-18
Delete asset tracking + publish asset copy

dheniges:
requires #423 

- On publish, copies image field assets to a new location on S3
- When deleting an image field, deletes the asset on S3 and decrements the resource tracker in the db


jjidt (commented):
Code looks good, can you check out image uploading in the web app, I'm not getting access to the uploaded image on this branch

dheniges:
I'm not able to reproduce your issue JJ. Tried in FF and Chrome. I did notice that the image component keeps the base64 data version of the image that's added, even after the image is successfully uploaded to the API, and the component doesn't render the CDN version until you refresh.

Tried a few different combos with published and non-published versions of an image field

jjidt (approved):
Seems like my issue was with using the pesky seed bonkey user. Works great!

PR #427, opened 2020-01-21
initial API rate limiting

dheniges:
- Adds the sliding-window-rate-limiter, which can use memory or redis store
- Adds kube config for running redis in the kube cluster
- Adds middleware to use rate limiter based on orgId, middleware used on all authenticated routes
- Returns 429 error code if rate limit exceeded

### DIFF
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -4517,6 +4517,11 @@
         "shimmer": "^1.1.0"
       }
     },
+    "cluster-key-slot": {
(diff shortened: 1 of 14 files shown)


jjidt (approved):
👍 Code looks good, still works locally. Gonna let the rate limiting get tested in the wild

jjidt:
also tested locally, and it gave me all those sweet sweet 429s

PR #429, opened 2020-01-22
bounce and complaint notifications via ses

dheniges:
Not ideal since these emails will only go to me, but at least we'll know if these rates spike


jjidt (approved):
+2 - Bonus +1 for these emails getting sent to :shockandawe:

PR #430, opened 2020-01-22
Fixes DND; Updates seedData

jcontonio:
* Fixes the bug where you could still drag/drop fields without being in edit mode
* Updates seed data so we have useful data for testing styles
* There are issues in the rendering, because our editor fucking sucks


dheniges (dismissed):
Nothing is draggable outside of edit mode except for image fields. I could still drag and reposition image fields

jjidt (approved):
👍 Drags, also drops

PR #432, opened 2020-01-22
using API logging code on web

dheniges:
Not very DRY, but why reinvent the wheel?

Allows us to look at web server performance, and add logging for specific web server events in the future.

To test, add the env vars in 1pass for the web .env. You'll need to `make installweb` as well


jjidt (approved):
👍 works! Also, don't mind the WETness here, most of it is config

PR #439, opened 2020-01-23
Mobile updates

jcontonio:
* Updates mobile views throughout the site
* Moves the "upgrade" notices out of the top header into its own subhead on mobile
* Makes sure we can have a home screen app that looks good

image


jjidt:
I get stuck when going to create an initial document on iphone 6s:

IMG_1603

jjidt:
IMG_B853C214A0A3-1

Delete icon is overlapping user reinvite link

dheniges:
0EDC89F4-1ABF-4359-9374-84968B259B6C_1_105_c

The placeholder text for strings and numbers gets cutoff. And what JJ said about creating documents being broken.

Otherwise I think desktop is good, opened other issues as I found them

dheniges (approved):
Mobile!

PR #440, opened 2020-01-23
public assignments

jjidt:
allow project readers and contributors to list project users


jjidt:
Currently on release - if you assign a user to a project as a reader or contributor, they'll get a 404 in their project list view because they don't have perms to fetch project assignments(getting fetched to display user avatars). This gives them the perm.

dheniges (approved):
Well I can't test this without breaking everything, but I am good with the perms changes :)

PR #441, opened 2020-01-23
Public / Read-only project fixes

jcontonio:
* Marks a project as public in the project list
* Removes unnecessary loading of ProjectUsersContainer if the user can't see it
* Removes users query in the DocumentToolbar, which caused the 404 on a project


jjidt (approved):
This is a nice solution 👍

PR #443, opened 2020-01-24
Adding more loaders everywhere

jcontonio:
* Adds some nice loaders in places I noticed the UI was unresponsive when dev was going nuts
* Need some help with why the empty state for the documents list is flashing. It should either be empty and loaded, or not empty and loaded. Getting a quick flash between those two states on a throttled connection


dheniges:
There's a sizable conflict in HeaderComponent

jcontonio:
Fixed the conflicts

dheniges:
I'm not seeing any issues here with or without a throttled connection. If I go to a project, the documents list loads as expected without flashing or showing anything unexpected. Are you seeing the no documents icon and message about adding documents? I don't see that

jcontonio:
Yeah, I do see that if I throttle using Fast 3G in Chrome

dheniges (approved):
:+1:

PR #445, opened 2020-01-24
Fixing font consistency

jcontonio:
Fixes Chrome / FF inconsistency, because Chrome was doing god knows what or why


dheniges (approved):
Looks good on Chrome/Safari/FF

PR #447, opened 2020-01-24
Dropdown fix

jcontonio:
Fixes dropdown menus like ProjectRole and OrgRole - Stops an infinite loop


dheniges (approved):
:clap:

I didn't realize you could do this with useRef, very handy. I only knew its usage in conjunction with the ref prop

PR #449, opened 2020-01-24
Fixes new document on mobile

jcontonio:
Added check for new document param


dheniges (approved):
This works but convinces me we do want a "+" button or something to add the doc. Clicking "done" on the ios keyboard doesn't create the doc, you have to click something else on the page to create the doc, which I don't think is intuitive

PR #450, opened 2020-01-24
clear out user project list when new project created

jjidt:
when a new project is created, clears out user projects and loading state so that they are refetched in the admin section


dheniges (approved):
:+1:

PR #451, opened 2020-01-24
Fixes the react flush warning

jcontonio:
Fixes the Form `Cannot flush updates when React is already rendering` warning message when submitting forms, like the Project Settings Info form


dheniges (approved):
:+1: form works and warning is gone

PR #452, opened 2020-01-25
Fixing keys list and endpoints

jcontonio:
This wasn't just on the iPad. Regular Safari and other browsers looked dumb once a key was selected

image


dheniges (changes requested):
Something is off, there's a white box beneath each endpoint:
Screen Shot 2020-01-27 at 8 22 47 AM

dheniges (approved):

PR #455, opened 2020-01-25
Touch polyfill for drag

jcontonio:
* Supports drag and drop on touch devices

### DIFF
--- a/web/src/client/js/components/App/App.js
+++ b/web/src/client/js/components/App/App.js
@@ -3,6 +3,11 @@ import PropTypes from 'prop-types'
 
 import 'CSS/app.scss'
 
+// https://github.com/Bernardo-Castilho/dragdroptouch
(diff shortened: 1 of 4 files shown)


dheniges:
Well this kind of works, but it's not too hard to get into broken states. Sometimes dragging things requires touching and holding the drag handler, sometimes you can just touch the field and start dragging it. Sometimes you touch and hold the field and it will start dragging.

So you can get into a situation where you drag a field, then you can't scroll because it just starts dragging fields.

I dunno if this is better or worse than not being able to drag fields at all. If [this bug](https://app.asana.com/0/1103604360526176/1159001815352817) gets fixed it might be OK. In that scenario you'd at least be able to exit drag mode if it got all wonky and resume normal doc editing

dheniges (approved):
Seems to work pretty well now 👍

PR #459, opened 2020-01-27
Adding Dirks button

jcontonio:
Butts 🍑


dheniges (changes requested):
Functionally feels solid. The hover state on the green "+" button removing the color seems inconsistent with all our other colored buttons. I would expect it to fade the green color or something, like the red delete field icons.

jcontonio:
Updated

dheniges (approved):
🍑

PR #460, opened 2020-01-27
Fixing draggable on document mode change

jcontonio:
This may help with the touch stuff too. Will merge release into that and test once this gets merged.


dheniges (changes requested):
This fixes the issue with the data fields (string, number) but image fields are still drag'n'droppable. I dunno if that's a different issue though?

jcontonio:
I think images are getting their regular drag and drop behavior from the OS. I will figure that out on this PR, but it is a separate issue.

jcontonio:
Ok, this should finally be fixed.

dheniges (approved):
💯

PR #463, opened 2020-01-29
help page updates

dheniges:
- New feedback@bonkeybong.com email for areas (like help page) where we're soliciting feedback and don't want them to get emails back from zendesk with a support ticket
- Only 1 API documentation link, since that's all we'll have live for now
- Wording updates to keep it positive


jcontonio (approved):
That’s much nicer. Nice work

PR #468, opened 2020-01-29
Updating project team menu

jcontonio:
Almost fixes it all, I need help with one thing.

1. This will load the first page of users by default in the menu
2. Start typing a name you know doesn't exist (zz)
3. Now delete the (zz) and see it goes back to the initial list
4. Now type "b" and see the Baddys and Biddys
5. Now delete the b, and you'll see that the searchResults still have the Baddys and Biddys loaded, but they shouldn't be... because the setState() in the container should have reset userSearch results

Is this a problem with passing null as a value to dataMapper?

### DIFF
--- a/web/src/client/css/inc/_react-select.scss
+++ b/web/src/client/css/inc/_react-select.scss
@@ -82,9 +82,6 @@ form {
         outline: none;
       }
     }
-    &__indicators {
-      display: none;
-    }
     &__placeholder {
       color: var(--react-select-placeholder-color);
       + div {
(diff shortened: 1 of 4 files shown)


jjidt:
After trying to track this problem through dataMapper/useDataService, I think I found a simpler solution in the component itself. Now, always sets the user search string in state, even if empty, and conditionally renders the initial page based on the search string rather than search results

dheniges (commented):
I think this is overall a better experience, but the issue Jay described is still present unless you delete the search result string.

Steps:
- Enter a search term
Do 1 of: 
- click the "X" to clear the search results
- click outside the menu

The search results still show up even though there's no search string present in the select menu

dheniges (approved):
🎉

PR #469, opened 2020-01-30
Org deletion job setup

jjidt:
- sets up endpoints for canceled org deletion
- moves admin auth to header `Authorization: Admin <adminsecret>`
- sets up kubernetes cronjob

### DIFF
--- a/kubernetes/jobs/configMaps.yaml
+++ b/kubernetes/jobs/configMaps.yaml
@@ -9,11 +9,4 @@ items:
   metadata:
     name: audit-log-backup-config
     namespace: default
-- apiVersion: v1
-  data:
-    delete.api_url: 'https://api.portway.app'
-  kind: ConfigMap
-  metadata:
-    name: deleted-org-cleanup-config
-    namespace: default
 kind: List
--- a/kubernetes/jobs/configMaps.yaml
+++ b/kubernetes/jobs/configMaps.yaml
@@ -9,4 +9,11 @@ items:
   metadata:
     name: audit-log-backup-config
     namespace: default
+- apiVersion: v1
+  data:
+    delete.api_url: 'https://api.portway.app'
+  kind: ConfigMap
+  metadata:
+    name: deleted-org-cleanup-config
+    namespace: default
 kind: List
--- a/kubernetes/jobs/deletedOrgCleanup.yaml
+++ b/kubernetes/jobs/deletedOrgCleanup.yaml
@@ -18,16 +18,11 @@ spec:
             env:
             - name: ADMIN_API_KEY
               valueFrom:
-                secretKeyRef:
-                  name: deleted-org-cleanup
-                  key: admin_api_key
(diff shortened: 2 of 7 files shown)


kubernetes/jobs/configMaps.yaml
```diff
@@ -9,4 +9,11 @@ items:
   metadata:
     name: audit-log-backup-config
     namespace: default
+- apiVersion: v1
+  data:
+    delete.api_url: 'https://api.portway.app'
```
dheniges: I would re-use the existing config map that already has this url. I think it's in the web config map, which arguably should be renamed/moved into its own config map, but I wouldn't worry about that for now. So you can drop the changes to this file

kubernetes/jobs/deletedOrgCleanup.yaml
```diff
@@ -0,0 +1,33 @@
...
+          restartPolicy: OnFailure
+          containers:
+          - name: deleted-org-cleanup
+            image: bonkeybong/portway_deleted_org_cleanup
+            env:
+            - name: ADMIN_API_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: deleted-org-cleanup
+                  key: admin_api_key
```
dheniges: We already have the admin key used in a secret, which you can see in the api deployment. Re-use that secret here. (note you can keep the `ADMIN_API_KEY` env var name and define it as the `admin_secret_key`)

kubernetes/jobs/deletedOrgCleanup.yaml
```diff
@@ -0,0 +1,33 @@
...
+            - name: LOG_ID
+              valueFrom:
+                secretKeyRef:
+                  name: deleted-org-cleanup
+                  key: log_id
+            - name: API_URL
+              valueFrom:
+                configMapKeyRef:
+                  name: deleted-org-cleanup-config
+                  key: delete.api_url
```
dheniges: this is where you can change the definition to pickup the web config map's public api url key

kubernetes/jobs/deletedOrgCleanup.yaml
```diff
@@ -0,0 +1,33 @@
...
+          containers:
+          - name: deleted-org-cleanup
+            image: bonkeybong/portway_deleted_org_cleanup
+            env:
+            - name: ADMIN_API_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: deleted-org-cleanup
+                  key: admin_api_key
+            - name: LOG_ID
```
dheniges: doesn't look like LOG_ID is used in the job, you can delete this env var

dheniges (changes requested):
Feedback on the kube config

jjidt:
updated

dheniges (approved):
Gets canceled orgs, can't delete canceled ones less than 30 days old, or non-canceled orgs  👍

PR #470, opened 2020-01-31
More launch mobile

jcontonio:
* Making navigation on panel content much nicer for mobile (admin, project settings)
* Makes bottom panel nav consistent with other iOS apps when in standalone mode
* Fixes project teams on mobile
* Fixes payment info on mobile
* Fixes admin users on mobile


dheniges (approved):
:+1:

PR #474, opened 2020-02-04
slack events

jjidt:
- signup
- cancel
- successfully charged


dheniges (changes requested):
Looks good, just add the SLACK_WEBHOOK_URL to the env var documentation. And then to the k8s config as a secret. Lemme know if you want a hand with that

jjidt:
I think I have it all in there now, check it out at your leisure

dheniges (approved):
Looks good

PR #475, opened 2020-02-04
Adding orientation detect

jcontonio:
1. Go to a project > document that is long and scrolly
2. Rotate your phone. The bottom toolbar should always be in the right spot

There is a delay in placing the toolbar back on rotate


jjidt (approved):
looks good, fixes the toolbar realignment issue 👍

PR #479, opened 2020-02-04
Outline view

jcontonio:
* Adds outline view, fixes editing mode issues
* Allows renaming of text bodies


jjidt (approved):
Initially seemed weird to have content disappear in edit mode, but it will be a lot easier to work with re-ordering docs with lengthy body fields. 

I like it/ works great 👍

PR #489, opened 2020-02-05
copy structuredValue for text fields on publish

dheniges:
only draft fields were getting structured values :idk:


jjidt (approved):
It works 👍

PR #493, opened 2020-02-06
info logging for bad password on login

jjidt:
- adds ability to pass logLevel on any error thrown in api
- passes info level for incorrect password


jcontonio (approved):
Looks good to me. Get a handled error in the logs

jjidt:
cool, I double checked with throwing 500 and 400 level errors and they are still getting the correct log level

PR #498, opened 2020-02-10
bump account invite token expiration to 14 days

dheniges:
Now free accounts have 14 days to click the email link before the link expires.


jcontonio (approved):
/shrug

PR #499, opened 2020-02-10
Fix image errors

dheniges:
Empty image fields (ie an image field with no uploaded image) were causing errors on delete and publish. Fixed.


jcontonio (approved):
Deleted and published images fine

PR #500, opened 2020-02-10
Removing styles

jcontonio:
Removing any nice markdown formatting styles because it just ends up breaking stuff


dheniges (changes requested):
The blockquote styling gets applied to the next line that's not a blockquote

Screen Shot 2020-02-10 at 4 52 09 PM

jcontonio:
This will require this PR to be merged: https://github.com/BonkeyBong/CodeMirror/pull/1

Then you'll need to do an `npm install` in web to get the correct Bonkey version of CodeMirror.

dheniges (commented):
Screen Shot 2020-02-21 at 10 21 55 AM

The problem is still there, the only way to get it to not apply the blockquote styling is to have a blank line between the last blockquote and the next line. It looks like a codemirror class getting applied to the line (`cm-quote`), so I dunno if there's anything to be done?

dheniges:
also just confirmed the quote class is added in the older codemirror version too, it just wasn't styled gray so you couldn't tell in the UI

dheniges (approved):

PR #506, opened 2020-02-11
Removing opacity change for disabled document

jcontonio:
When publishing, or moving fields between documents we had an opacity change on the document. Removed that.


dheniges (approved):
🔢

PR #509, opened 2020-02-11
Updating language for expiration in beta email

jcontonio:
Uses the `accountInviteExpiration` value from tokenSettings. Not sure if that's cool?


dheniges (changes requested):
This doesn't actually pull in the expire setting, I think because tokensettings is exporting a single object and doesn't have named exports

dheniges (approved):
the email works now, shows the 14 days and the sentence reads nicely

PR #511, opened 2020-02-12
Naming field with file name

jcontonio:


dheniges (approved):
:+1: works to go into the feature branch, but still need to figure out the hard part of this feature, ie not overwriting a field name someone has explicitly entered

PR #512, opened 2020-02-13
Outline panel

jcontonio:
* Adds outline panel instead of switching "modes"
* Requires the update fields bug to re-render the document when fields change

### DIFF
--- a/web/src/client/js/components/DocumentFields/DocumentFieldsComponent.js
+++ b/web/src/client/js/components/DocumentFields/DocumentFieldsComponent.js
@@ -2,7 +2,7 @@ import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 
-import { DOCUMENT_MODE, FIELD_TYPES } from 'Shared/constants'
+import { FIELD_TYPES } from 'Shared/constants'
 import DocumentFieldComponent from './DocumentFieldComponent'
 import FieldTextComponent from 'Components/FieldText/FieldTextComponent'
 import FieldNumberComponent from 'Components/FieldNumber/FieldNumberComponent'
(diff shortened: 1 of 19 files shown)


web/src/client/js/components/DocumentFields/DocumentFieldsComponent.js
```diff
@@ -24,6 +24,8 @@ const DocumentFieldsComponent = ({
 }) => {
   const [settingsForField, setSettingsForField] = useState(null)
 
+  console.log(fields)
```
dheniges: 👀

dheniges (changes requested):
Documents with zero or one field (I think) error out:
Screen Shot 2020-02-13 at 1 43 28 PM
Screen Shot 2020-02-13 at 1 43 17 PM

dheniges (commented):
I moved this PR into a feature branch, there's a lot going on. It's probably easiest to merge this into the feature branch now. I reviewed the code but a bunch that needs to be addressed before it's releasable.

Overall I think this is a good approach, it lets people see what they're changing in a format that works with drag and drop. I think it will be the best experience of anything we've had so far. I'll check it out on mobile after it's in a more stable spot.

Issues:
- Can't double click or highlight to select the field names in the outline
- Drag/drop event handler issues. Drag a field around a couple times. Sometimes it updates in the doc, sometimes it doesn't
- Why is there a "Save changes" button? All the actions save when performed (re-order fields, rename field, delete field)

Other feedback:
- What do you think about an active field that's highlighted in both the outline and the main content document? So if I click on a field it's obvious which one I'm editing. I think this would help address the issue of not knowing which outline field corresponds to the main content area field. Examples: text fields back to back, small height fields at the bottom of the doc, etc. Seems like that would compliment the current scroll to field behavior which works great.

jcontonio:
> Can't double click or highlight to select the field names in the outline

You think there should be an "edit" type icon/button?

> Why is there a "Save changes" button? All the actions save when performed (re-order fields, rename field, delete field)

I added this for users who may not realize that changes are saved all the time, and may not see the "Done" button. Do you think an "X" type button would work instead?

PR #514, opened 2020-02-13
Add warning to project teams

jcontonio:
* Adds a warning on project settings team page, if the org doesn't have any users added
* Also fixes alignment issues on Admin Notices for trial expiration title

Screen Shot 2020-02-13 at 3 56 33 PM

### DIFF
--- a/web/src/client/js/components/ProjectSettings/ProjectSettingsTeams/ProjectSettingsTeamsComponent.js
+++ b/web/src/client/js/components/ProjectSettings/ProjectSettingsTeams/ProjectSettingsTeamsComponent.js
@@ -28,7 +28,7 @@ const ProjectSettingsTeamsComponent = ({
   const [newUserRole, setNewUserRole] = useState(PROJECT_ROLE_IDS.READER)
 
   // Show a message if there are no users other than the project owner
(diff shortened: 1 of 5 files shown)


web/src/client/js/components/ProjectSettings/ProjectSettingsTeams/ProjectSettingsTeamsComponent.js
```diff
@@ -25,6 +26,7 @@ const ProjectSettingsTeamsComponent = ({
   const selectRef = useRef()
   const [newUserId, setNewUserId] = useState(null)
   const [newUserRole, setNewUserRole] = useState(PROJECT_ROLE_IDS.READER)
+  const showAddUsersMessage = users.length === 0
```
dheniges: Looks like the `users` here are the search results users with currently assigned users filtered out. So if you have more than one user, and all the users are assigned to the project already, the warning still shows up

dheniges (changes requested):
The expiration notice for the trial looks good

The check for org users needs updating. See review comment.

dheniges (changes requested):
This doesn't work, the message doesn't show up at all

dheniges (approved):
:+1:

PR #515, opened 2020-02-14
Changing icon theming

jcontonio:
You should now be able to toggle between dark mode and light mode and get the proper icon colors


dheniges (approved):
Looks good. I probably didn't find all the icons, but I found a lot and they all looked good :shockandawe:

PR #516, opened 2020-02-14
Hiding fullscreen button for mobile safari

jcontonio:
Full screen mode for dom elements is disabled in iOS < 13.2. We should hide the full screen button if the browser doesn't support fullscreen API, or if the browser is an iPad.

~iPadOS however, DOES support fullscreen, except you can't actually do anything in it (typing is disabled). Once you tap in the window in fullscreen, it collapses back. There's no way to hide the button on iPadOS because there's no way to detect proper functionality.~


dheniges (changes requested):
This disables the full screen button in chrome desktop

dheniges (approved):
Good on Mac Safari/FF/Chrome and good on mobile Safari :+1:

PR #517, opened 2020-02-14
portway sync

jjidt:
Adds `sync` server to handle all your user sync needs

Not yet configured in docker-compose, so you'll need to start up the server manually:

- get a local redis server running on the default port (6379)
- copy example.env to a .env file
- navigate to `sync` directory and `npm run serve`

This is currently only handling document user rooms, no field stuff yet


jcontonio (approved):
Works and looks good 👍

PR #518, opened 2020-02-14
Feature: Image name / file name

jcontonio:
- [x] Uses filename on creation of field
- [x] Doesn't rename the field if the field was given a custom name


dheniges:
Per discussion, if we can get this going so the client only assigns the image file name when it's a new field

dheniges (approved):
Perfect

PR #519, opened 2020-02-14
Feature: Outline panel

jcontonio:
- [x] Bug: dragging/dropping after a while doesn't re-order the document
- [x] Bug: editing content in the document causes a re-render after save, whether in regular or outline mode. It will push the cursor back to the top text field if you have multiple fields, and flashes the whole editor
- [x] Basic outline
- [x] Bug: Figure out edit / double-click / selecting of field name
- [x] UI: Maybe a different close / save bit here


dheniges (approved):
This. Is. Outline Mode.

:tada: :taco: :dumps: :shockandawe: :soundguyspissed: 🍌 🌵 🎈 🎉

PR #520, opened 2020-02-14
Fixed header title when navigating back to projects list

jcontonio:
If you went to a project, and then back to the project list, the header never changed back to "Projects"

### DIFF
--- a/web/src/client/js/components/Navigator/NavigatorContainer.js
+++ b/web/src/client/js/components/Navigator/NavigatorContainer.js
@@ -9,6 +9,7 @@ import {
   ORGANIZATION_ROLE_IDS,
   ORGANIZATION_SETTINGS,
   PATH_PROJECT,
+  PATH_PROJECTS,
   PATH_PROJECT_CREATE,
   PRODUCT_NAME,
   PROJECT_ROLE_IDS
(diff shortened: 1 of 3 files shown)


web/src/client/js/libs/currentResource.js
```diff
@@ -28,7 +28,15 @@ const currentResource = (resourceName, path) => {
     case 'project':
       // Specifically look at project here, as we may be looking at documents
       const projectPathArray = path.split('/')
+
+      // We're back on the projects list
+      if (projectPathArray[1] === 'projects') {
```
dheniges: is there a constant that can be used for this path? It would help prevent breaking this again in the future

web/src/client/js/components/Navigator/NavigatorContainer.js
```diff
@@ -35,6 +36,13 @@ const NavigatorContainer = ({ history, location }) => {
     currentResource('project', location.pathname), [location.pathname]
   )
 
+  let title = ''
+  if (location.pathname === PATH_PROJECTS) {
+    title = 'Projects'
+  } else {
+    title = project ? project.name : 'Projects'
+  }
```
dheniges: Can simplify and only use the "Projects" string once:
```
let title = 'Projects'
if (location.pathname !== PATH_PROJECTS && project) {
  title = project.name
}
```

dheniges (changes requested):
Couple code comments, plus another related bug:
The "Help" page is using the project selector thingy and will still show the last project you were looking at instead of "Projects"

dheniges (approved):
This looks good, I think having the titles in one spot will make this easier to maintain

PR #521, opened 2020-02-14
Fixed user validation

jcontonio:
* Fixed API responses sending publicMessage instead of message (the original bug)
* Updated the Add User form to use the standard Form / Validation fields like every other form in the app
* Test adding, adding an existing user, canceling, canceling after getting validation errors

### DIFF
--- a/web/src/client/js/components/Admin/AdminUsers/AdminUsersCreateForm.js
+++ b/web/src/client/js/components/Admin/AdminUsers/AdminUsersCreateForm.js
@@ -3,20 +3,25 @@ import PropTypes from 'prop-types'
 
 import { debounce } from 'Shared/utilities'
 import { ORGANIZATION_ROLE_IDS } from 'Shared/constants'
-import SpinnerContainer from 'Components/Spinner/SpinnerContainer'
+
+import Form from 'Components/Form/Form'
 import FormField from 'Components/Form/FormField'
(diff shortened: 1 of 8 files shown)


web/src/client/js/components/Admin/AdminUsers/AdminUsersCreateForm.js
```diff
@@ -3,20 +3,25 @@ import PropTypes from 'prop-types'
...
+const AdminUsersCreateForm = ({ cancelHandler, disabled, errors, formId, submitHandler }) => {
   const [name, setName] = useState()
   const [email, setEmail] = useState()
   const [orgRole, setOrgRole] = useState(ORGANIZATION_ROLE_IDS.USER)
 
-  const borderColor = getComputedStyle(document.documentElement).getPropertyValue('--theme-border')
+  function getEmailOrUsersValidationErrors() {
+    // const emailErrorsArray = errors.email ? errors.email : []
+    // const usersErrorsArray = errors.users ? errors.users : []
+    // return [...emailErrorsArray, ...usersErrorsArray]
```
dheniges: deleteme

dheniges (changes requested):
Works great, just small comment on commented code

dheniges (approved):
👍 :sealeel: :dumps:

PR #522, opened 2020-02-14
Fixed text selection in outline item

jcontonio:


dheniges (approved):
Sweet, this works nicely

PR #523, opened 2020-02-14
Fixed drag operations

jcontonio:
Fixed drag operations
Changed outline toggle UI


dheniges (approved):
Non-dragging changes are kewl.

But, these changes don't seem any better or worse for the dragging. Still have numerous unexpected drag'n'drops that don't work. You can see my screenshot should have a text area between the two numbers that doesn't. No errors in the console or anything.

Screen Shot 2020-02-14 at 4 46 23 PM

PR #524, opened 2020-02-14
Intro project for new accounts

dheniges:
Copies a prod Portway project to new orgs to give them help and advice

Needs:
- [x] add owner as project admin
- [x] add k8s configs for Portway API access to project


jcontonio (commented):
Not getting the whole document. [This document](https://portway.app/d/project/21/document/43) is published, but I am only getting the first two fields.

Screen Shot 2020-02-17 at 4 25 22 PM

api/src/coordinators/signUp.js
```diff
@@ -38,6 +39,11 @@ async function createUserAndOrganization(name, email) {
     stripeId: customer.id
   })
 
+  // Letting this happen in the background. Should occur before user verifies their email
+  // and if there's an error we don't want it to prevent signup. Note this has to occur after
+  // the ownerId has been set on the org!
+  introCoordinator.copyIntroProjectToOrg(organization.id)
```
dheniges, line 45: @jcontonio give this another shot, I fixed a race condition but nothing with the images. If you don't see an error in the API logs because it's too far back, add an `await` here

dheniges:
@jcontonio see my comment on the await, but also note I merged release so to eliminate other issues run the app with `docker-compose up --build` to make sure you get the new node 12.16.0 images with all the required packages

jcontonio (approved):
I get a project with the full document 👍

PR #525, opened 2020-02-15
docker/docker-compose config for sync server

jjidt:


sync/Dockerfile
```diff
@@ -0,0 +1,21 @@
+
+FROM node:12.13.0-alpine
+
+# bcrypt and alpine linux have some issues:
+# https://github.com/kelektiv/node.bcrypt.js/wiki/Installation-Instructions#alpine-linux-based-images
+RUN apk --no-cache add --virtual builds-deps build-base python
```
dheniges: I don't think you need these lines?

sync/Dockerfile-prod
```diff
@@ -0,0 +1,29 @@
+
+FROM node:12.13.0-alpine
+
+# bcrypt and alpine linux have some issues:
+# https://github.com/kelektiv/node.bcrypt.js/wiki/Installation-Instructions#alpine-linux-based-images
+RUN apk --no-cache add --virtual builds-deps build-base python
```
dheniges: Same. I don't see bcrypt in sync which is why I had to add these dependencies

dheniges (changes requested):
Couple things can be removed from the docker images -- see comments. Otherwise looks good

jjidt:
updated

dheniges (approved):
still builds 👍

PR #526, opened 2020-02-17
Fixed document mounting/unmounting on edit

jcontonio:
The loading status of fields triggered the flash. We don't really need it


dheniges (approved):
Yep, solves the problem 👍

PR #528, opened 2020-02-18
Fixing drop handling

jcontonio:
Dropping an item when your cursor is *just* outside of the LI's wouldn't trigger an onDrop, because you were technically dropping on the OL. This solves that issue.


dheniges (approved):
:tada: :shockandawe:

PR #530, opened 2020-02-18
Added confirmation theming

jcontonio:
All confirmations were getting red primary buttons. This gives us the ability to theme that button, depending on the action. Actions like Publishing and Upgrading plans shouldn't be "red" since that's more of a danger color.


dheniges (approved):
I didn't check them all, but checked a bunch

PR #532, opened 2020-02-19
Feature FTE (First Time Experience or Full Time Employee, your choice)

dheniges:
- [x] Portway read secret for intro project has been added to dev and prod
- [x] Intro project has special flag for UI 
- [x] New dashboard component with intro project highlight


api/src/coordinators/field.js
```diff
@@ -11,6 +21,32 @@ const addFieldToDocument = async function(documentId, body, file) {
...
+  // Not ideal to be passing the multer object around, but that's a larger rewrite to fix
+  const file = {
+    originalname: name,
+    mimetype: lookup(url),
+    path: filePath,
+    size: fileStats.size
+  }
+
+  const result = await fieldCoordinator.addFieldToDocument(documentId, body, file)
+  fs.unlink(filePath, () => {}) // don't need to await
```
jjidt: does this need to be wrapped in a try/catch to log/ignore error? otherwise we could get additional express errors along with it

api/src/coordinators/signUp.js
```diff
@@ -38,6 +39,11 @@ async function createUserAndOrganization(name, email) {
     stripeId: customer.id
   })
 
+  // Letting this happen in the background. Should occur before user verifies their email
+  // and if there's an error we don't want it to prevent signup. Note this has to occur after
+  // the ownerId has been set on the org!
+  introCoordinator.copyIntroProjectToOrg(organization.id)
```
jjidt: same question here on the try/catch

jjidt (commented):
Just the one question about try/catching non-awaited async tasks, other than that looks/works great 💯

jjidt:
Actually, looks like the intro project doesn't show up until you create another project. If it's the only one it doesn't show up.

Screen Shot 2020-03-06 at 3 33 27 PM

jjidt (approved):
:shockandawe:

PR #534, opened 2020-02-19
field sync

jjidt:
- Adds field handling to sync server
- @jcontonio I need your help with this, I don't want to step on your md editor code ->
when a user updates a field, the event is successfully transmitted and data refetched by fellow room users, and even though the updated field is making it to the field component, the displayed field value isn't getting updated


jjidt:
have this updating field display as its changed by another user, we should thoroughly test this on dev to make sure its not jacking the cursor position as you're working

dheniges:
@jjidt merge conflict. womp womp

jjidt:
updated

dheniges (approved):
I dunno if this is perfect, but it's fun to play around with therefore 👍 

But seriously we should do some joint QA like you mentioned JJ, but I think this is good to go into the feature branch

PR #535, opened 2020-02-19
jwt sync auth

jjidt:
to test => set a different JWT_SECRET in sync .env file than you have in the api .env, it should stop you from connecting

can't pass data along the socket object like we do with express middleware req objects, so currently passing the userId as an additional query param rather than decoding it out of the jwt


dheniges (approved):
Spent some time combing through the sync stuff just to see what's going on. This works with the jwt auth, but seems like any valid jwt token will grant you access to any any "room" (or whatever the correct socket.io parlance is). Is that accurate so far?

This PR is good to merge though, just catching up on sync :idt:

jjidt:
@dheniges that's right ^ I'm thinking maybe we should namespace all the sync data in redis with the org id, and pull that id directly from the jwt on socket connection. That way they will only have access to creating/consuming data from their own org

dheniges:
I think that plan sounds perfect, as we previously discussed I don't mind exposing user ids in the same org if someone wants to get all wild

PR #536, opened 2020-02-21
jwt decoding/org namespacing for sync data

jjidt:


dheniges (approved):
V. nice, would org auth again

PR #541, opened 2020-02-24
Feature sync

jjidt:


dheniges:
Was just thinking on this, how difficult would it be to skip the socket connection for solo plans?

jcontonio (approved):
:jjmad: :pee: :dirk:

PR #542, opened 2020-02-24
fix require path to work on dev and prod

dheniges:
The require path was going out too many directories and had `src` hard-coded when the prod build puts files in `dist`. Thus an error when trying to start the API.


jjidt (approved):
👍

PR #545, opened 2020-02-25
Sync logging

jjidt:
more copy-pasta for logging, so maybe we should break that out into our own package

pretty straight forward, also adds some missing things to web example.env


dheniges (approved):
Logs :log:

PR #551, opened 2020-02-27
Notify via slack when someone submits support form

jjidt:
squishymuffins


dheniges (approved):
Looks good, kubes look good 👍

PR #554, opened 2020-02-28
join room on field events

jjidt:
Note that timeouts are set to 60 seconds currently to help testing, added a TODO to set to larger values


jcontonio (approved):
Nice!

PR #557, opened 2020-03-03
user fetch by batch ids

jjidt:


dheniges (approved):
Lovely, fetches the users

PR #562, opened 2020-03-03
First-time experience UI

jcontonio:
* Adds a new Dashboard component that contains a special project, and the projects list
* Fixes a bug in the Form disable logic


web/src/client/js/actions/organization.js
```diff
@@ -6,7 +6,11 @@ import { NOTIFICATION_TYPES, NOTIFICATION_RESOURCE } from 'Shared/constants'
 export const fetchOrganization = (orgId) => {
   return async (dispatch) => {
     dispatch(Organizations.requestOne(orgId))
-    const { data, status } = await fetch(`v1/organizations/${orgId}`)
+    // Get the org and the org's special project Id
+    const results = await Promise.all([fetch(`v1/organizations/${orgId}`), fetch(`v1/organizations/${orgId}/introProjectId`)])
```
dheniges, line 10: It seems like this action isn't hit if you go to the admin page for an organization. I wound up with the intro project not getting the special styling:
1. Go to the organization admin page
1. Change a setting on the organization (I checked "Everyone can create projects")
1. Click the Portway logo to go back to the dashboard. The intro project is not styled.

dheniges (changes requested):
There are definitely a few edge case scenarios where the special project doesn't get styled. Testing I ran into through another path too but I couldn't figure out the steps to repeat it. Probably another alternate org loading path.

Everything else seems to work with this that I tested.

Last thing is styling on the projects list now, I feel like it needs some divider so it's clear which user icons are associated with the project:
Screen Shot 2020-03-04 at 8 55 12 AM

jcontonio:
Fixed up, and have a surprise for you :dirk:

dheniges:
@jcontonio doesn't look like you pushed up any changes?

jcontonio:
Pushed

web/src/client/js/components/Dashboard/DashboardSpecialProject.js
```diff
@@ -0,0 +1,110 @@
...
+import OrgPlanPermission from 'Components/Permission/OrgPlanPermission'
+import ProjectPermission from 'Components/Permission/ProjectPermission'
+import Stars from './Stars'
+
+// Lazy loading because not all users have access to this
+const ProjectUsersContainer = lazy(() => import(/* webpackChunkName: 'ProjectUsersContainer' */ 'Components/ProjectUsers/ProjectUsersContainer'))
+
+import './_SpecialProject.scss'
+
+const lsShowMessage = localStorage.getItem('sp-message')
```
dheniges: What do you think about adding a date check to this? Like if the project was created > 30 days ago, never show the message?

web/src/client/js/components/Dashboard/DashboardSpecialProject.js
```diff
@@ -0,0 +1,110 @@
...
+
+const DashboardSpecialProject = ({ deleteHandler, project }) => {
+  const [showMessage, setShowMessage] = useState(lsShowMessage)
+  const assignedProject = project.accessLevel == null
```
dheniges: I'm not following this logic. A project can have a default accessLevel independent of users being individually assigned to the project

dheniges (changes requested):
A couple other comments, and there's a lot of whitespace at larger screen sizes, I think it needs to be tightened up or dividers added or something. The project users and icons are kinda floating in the abyss. The hover effect helps offset this when you actually interact with it, but see the screenshot:
Screen Shot 2020-03-04 at 3 51 30 PM

web/src/client/js/components/Dashboard/DashboardSpecialProject.js
jcontonio: We don't show the ProjectUsers list if the project has an accessLevel set, because it is accessible to everyone. That's when we show the "Unlocked". I think I just have a very terrible variable name

dheniges: ah gotchya, maybe just call it `showProjectUsers` ?

jcontonio:
Changed ProjectList to use the sortable Table component, and rendering the special project as the first item in that table if it exists

This is prepping for sortable / filterable projects

dheniges (approved):
Looks great 👍

PR #563, opened 2020-03-03
sync string and number fields

jjidt:
also adds number length limit

### DIFF
--- a/sync/src/app.js
+++ b/sync/src/app.js
@@ -10,7 +10,6 @@ app.use(logger('dev'))
 app.use(json())
 app.use(urlencoded({ extended: true }))
 app.use(cookieParser())
-
 // Mount a base router to respond to / requests
 // for verifying service is running at FQDN.
 // Added for letsencrypt verification of running service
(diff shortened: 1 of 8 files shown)


sync/src/app.js
```diff
@@ -10,7 +10,6 @@ app.use(logger('dev'))
 app.use(json())
 app.use(urlencoded({ extended: true }))
 app.use(cookieParser())
-
```
dheniges, line 13: It's a good change, it's a good change.
https://www.youtube.com/watch?v=Rt3gcSyL5vg

dheniges (changes requested):
It works in the update scenario, but some issues I came across. The 2nd and 3rd issues may be separate from this PR, in which case they can go in Asana.

- Two react warnings, see screenshots below
- When user 1 adds a new string or number, user 2 doesn't see it until user 1 enters a value
- Reordering fields isn't reflected for other users until refresh

Screen Shot 2020-03-04 at 8 20 06 AM
Screen Shot 2020-03-04 at 8 20 31 AM

jjidt:
It made sense to go ahead and fix sync for create/remove/reorder, check it :shockandawe:

dheniges (approved):
What can I say? It syncs :tada: :realjj:

PR #564, opened 2020-03-04
Adds clickable URLs to markdown fields

jcontonio:
Let's see if we like it this way for now, and if we want to change it we can


dheniges (commented):
Works cross-browser.

What do you think about this scenario? 
`[google](google.com)`
I would expect that to perform the same way as if I typed "google.com" in a browser. Instead it currently acts like a relative link and takes you to portway.app/documents/google.com

jcontonio:
Updated

dheniges (approved):
V. nice

PR #565, opened 2020-03-04
Sync UI: Who is in the document

jcontonio:
* Fetches missing users from state
* Displays the avatar or a backup for users that aren't the current user


dheniges (commented):
It works for the scenarios I ran through. The network requests look good, it's only requesting what it needs. We'll see if JJ has any comments, but looks like all the socket events are the same so probably good?

web/src/client/js/components/DocumentUsers/DocumentUsersComponent.js
```diff
@@ -1,10 +1,44 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
+import './_DocumentUsers.scss'
+
 const DocumentUsersComponent = ({ activeUsers }) => {
+  console.log(activeUsers)
```
jjidt: remove?

jjidt (changes requested):
looks good to me, tracking loading state through individual ids seems like the way to go 👍 just the console log, unless you want that in there for now

PR #566, opened 2020-03-04
createdBy when project is added

jjidt:
add createdBy to projects from requestor user id, this is an optional field and not directly tied to user model by foreign key. This way, these could be null, or tied to a user that no longer exists, but I don't think we necessarily care about that.


dheniges (approved):
Looks good, I like not tying it the user via schemas to simplify deletion scenarios etc. since it's not a required field

PR #567, opened 2020-03-04
Delete sub resources for projects and documents

dheniges:
- Project deletion will now delete associated docs
- Doc deletion will delete doc, fields, and versions


jjidt (approved):
👍 looks good, nice tests 💯

PR #569, opened 2020-03-05
field sync user icons

jjidt:


jcontonio:
I updated to reuse the same users component with some options. However, I see the current in the field.

jcontonio (changes requested):
See comment, forgot to review

jjidt:
updated with user filter

jcontonio (approved):
Nice 👍

PR #572, opened 2020-03-06
Offline message / warning

jcontonio:
* Adds an ApplicationContext to the whole app for `networkStatus`
* Reacts to network changes and notifies the user
* Will attempt to navigate back to the previous screen if loading a chunk that isn't in memory yet 

Screen Shot 2020-03-06 at 3 12 10 PM

Need help with something:

1. Go to projects screen
2. Set your network to Offline
3. Click a project
4. Clicking the "Return to previous screen"
5. See how that doesn't work... click it again though, and you'll see the view flash before it goes to your homepage or whatever.... is it because ErrorBoundary still thinks it has an error?


jcontonio:
Removed the "back" functionality. Ready for review

web/src/client/js/components/ErrorBoundary/ErrorBoundaryComponent.js
```diff
@@ -33,8 +47,11 @@ class ErrorBoundary extends React.Component {
...
+ErrorBoundary.contextType = ApplicationContext
+
 ErrorBoundary.propTypes = {
-  children: PropTypes.node
+  children: PropTypes.node,
+  history: PropTypes.object,
 }
 
-export default ErrorBoundary
+export default withRouter(ErrorBoundary)
```
dheniges: is withRouter no longer needed since the back thing doesn't work?

web/src/client/js/components/ErrorBoundary/ErrorBoundaryComponent.js
```diff
@@ -33,8 +47,11 @@ class ErrorBoundary extends React.Component {
   }
 }
 
+ErrorBoundary.contextType = ApplicationContext
+
 ErrorBoundary.propTypes = {
-  children: PropTypes.node
+  children: PropTypes.node,
+  history: PropTypes.object,
```
dheniges: no longer used?

dheniges (changes requested):
Works nicely, I think just some leftover withRouter/history

dheniges (approved):
:bassface:

PR #573, opened 2020-03-07
sync connection

jjidt:
- remove connection error message, just don't do any sync things until we have a connection
- also fixes proptype error


jcontonio (approved):
Works!

PR #574, opened 2020-03-07
FTE: Checking for specialProject

jcontonio:
This *should* show the specialProject if it's the only project, but waiting for DNS issues to make sure.


dheniges (approved):
:+1: just confirmed this fixes the problem

PR #575, opened 2020-03-09
FieldText: Hide textareas and fix autofocus of new fields

jcontonio:
* Added back the functionality to focus new body fields automatically
* Clears lastCreatedFieldId in more scenarios
* Hides textareas from flashing unstyled in Dirk's crazy Twitter documents (hopefully)


dheniges (approved):
Didn't test on crazy twitter docs, but everything thing else looks good :dumps:

PR #579, opened 2020-03-09
Fixing field image settings height

jcontonio:
When you have a short image, fixes the height on image settings


dheniges (approved):
:dirk:

PR #583, opened 2020-03-10
Removing contributor from project settings

jcontonio:
In project toolbar, project contributors shouldn't get access to project settings


dheniges (approved):
:haha:

PR #585, opened 2020-03-10
Fixing users list padding when in readOnly mode

jcontonio:
Fixes this spacing issue when in a readOnly doc and there are users in the doc

https://app.asana.com/0/1103604360526176/1165666014843865


dheniges (approved):
:loiterer:

PR #587, opened 2020-03-11
Date field

dheniges:
- Adds date field on API
- Functional through API
- Needs doc updates and web updates to support


jjidt (approved):
Nice 👍

PR #588, opened 2020-03-11
Update help section

jcontonio:
* Adds form to help section
* Removes "preview" copy
* Adds check for app vs website form submission in the endpoint

Adds two new web env vars. Production / Dev should get the following

```
SUPPORT_FORM_SUBMIT_ORIGIN=https://getportway.com
SUPPORT_FORM_REDIRECT_TO=https://getportway.com/support?received=true
```

### DIFF
--- a/web/src/client/js/sections/Help/HelpSection.js
+++ b/web/src/client/js/sections/Help/HelpSection.js
@@ -40,12 +40,14 @@ const HelpSection = () => {
             </div>
             <aside className="help-section__aside">
               <h2>Documentation</h2>
-              <ul className="list list--blank">
+              <ul className="list list--blank help-section__descriptive-list">
                 <li>
-                  <a href={URL_GUIDES} target="_blank" rel="noopener noreferrer">Guides</a>
+                  <p>Learn what you can do with Portway by perusing our handy guides website.</p>
+                  <a href={URL_GUIDES} className="btn btn--small btn--white" target="_blank" rel="noopener noreferrer">View the Guides</a>
                 </li>
                 <li>
-                  <a href={URL_API_DOCS} target="_blank" rel="noopener noreferrer">API Documentation</a>
+                  <p>Develop your own applications, integrations, or shortcuts using the robust Portway API.</p>
+                  <a href={URL_API_DOCS} className="btn btn--small btn--white" target="_blank" rel="noopener noreferrer">Read the API</a>
                 </li>
               </ul>
               <h2>Helpful links</h2>
--- a/web/src/client/js/sections/Help/HelpSection.js
+++ b/web/src/client/js/sections/Help/HelpSection.js
@@ -1,8 +1,15 @@
 import React, { useEffect } from 'react'
 import { Helmet } from 'react-helmet'
 
-import { URL_API_DOCS, PRODUCT_NAME, FEEDBACK_EMAIL } from 'Shared/constants'
+import {
+  URL_API_DOCS,
+  URL_GUIDES,
+  URL_PRIVACY,
+  URL_TWITTER,
+  PRODUCT_NAME,
+} from 'Shared/constants'
 
+import HelpForm from './HelpForm'
 import './_Help.scss'
 
 const HelpSection = () => {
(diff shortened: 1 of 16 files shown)


web/src/server/controllers/support.js
```diff
@@ -8,7 +8,13 @@ const SupportController = function(router) {
       `:question: Support :question: \n Email: ${email} \n Name: ${name} \n Company: ${company} \n Subject: ${subject} \n Message: ${message}`
     )
     await addRequest(email, message, company, name, subject)
-    res.redirect('https://getportway.com/support?received=true')
+    if (req.get('origin') === 'https://getportway.com') {
+      res.redirect('https://getportway.com/support?received=true')
+    } else if (req.get('origin') === 'http://localhost:8080') {
+      res.redirect('http://localhost:8080/support?received=true')
```
dheniges: do we want this on prod? Could move to an env var like `SUPPORT_FORM_SUBMIT_ORIGIN` for different values locally and on prod

web/src/client/js/sections/Help/HelpSection.js
```diff
@@ -13,68 +20,43 @@ const HelpSection = () => {
...
-              </p>
-
-              <p className="section">
-                <b>Thanks again!</b>
-                <br />
-                Your friends at BonkeyBong ❤️
+                Have a feature request or having a problem? Let us know, and we will try our hardest
+                to get back to you within a day. Don’t worry, we don’t do anything creepy with your
+                information. If you’d like to know more about that, read our{' '}
+                <a href={URL_PRIVACY} target="_blank" rel="noopener noreferrer">privacy policy</a>.
```
dheniges: I think since users are already logged in we don't need to call out the creepy privacy thing here, or make it more subtle. I like the link in the sidebar.

Also, I changed the copy for the "get back to you within a day" on the marketing site because I likely will not try my hardest to respond in a day. I will try to respond in a timely fashion. This is what marketing says now for the form:
```
        <p>
          Have a feature request or having a problem? Let us know, and we will get back to you as soon as we can.
        </p>
        <p>
          We’d love to hear from you.
        </p>
```

Thoughts?

web/src/client/js/sections/Help/HelpSection.js
```diff
@@ -13,68 +20,43 @@ const HelpSection = () => {
...
             <aside className="help-section__aside">
+              <h2>Documentation</h2>
+              <ul className="list list--blank help-section__descriptive-list">
+                <li>
+                  <p>Learn what you can do with Portway by perusing our handy guides website.</p>
+                  <a href={URL_GUIDES} className="btn btn--small btn--white" target="_blank" rel="noopener noreferrer">View the Guides</a>
+                </li>
+                <li>
+                  <p>Develop your own applications, integrations, or shortcuts using the robust Portway API.</p>
+                  <a href={URL_API_DOCS} className="btn btn--small btn--white" target="_blank" rel="noopener noreferrer">Read the API</a>
```
dheniges: should say "Read the API Docs"

dheniges (changes requested):
Some copy thoughts etc.

web/src/server/controllers/support.js
```diff
@@ -8,11 +8,11 @@ const SupportController = function(router) {
       `:question: Support :question: \n Email: ${email} \n Name: ${name} \n Company: ${company} \n Subject: ${subject} \n Message: ${message}`
     )
     await addRequest(email, message, company, name, subject)
-    if (req.get('origin') === 'https://getportway.com') {
-      res.redirect('https://getportway.com/support?received=true')
-    } else if (req.get('origin') === 'http://localhost:8080') {
-      res.redirect('http://localhost:8080/support?received=true')
+    // If we're getting hit from the external website, do a redirect
+    if (req.get('origin') === process.env.SUPPORT_FORM_SUBMIT_ORIGIN) {
+      res.redirect(process.env.SUPPORT_FORM_REDIRECT_TO)
```
dheniges: this doesn't need to be a separate env var because the path doesn't change between environments. It can be a path constant that's appended to the `SUPPORT_FORM_SUBMIT_ORIGIN`

Screen Shot 2020-03-13 at 8 14 55 AM

dheniges (changes requested):
Text updates look good, but only need 1 env var

dheniges (approved):
:+1:

PR #592, opened 2020-03-17
Remote change acceptance UI

jjidt:


dheniges (commented):
This seems good to merge. Things I noticed:
1) console warning from the string component:
```
Failed prop type: The prop `isCurrentlyFocusedField` is marked as required in `FieldStringComponent`, but its value is `undefined`.
```
2) User's name that made the change presumably needs to be added
3) Whatever we do with competing changes back and forth

jjidt:
updated, keeping the remote changes (and user name) around when they focused a different field was a little tricky, so check out the functionality of that

dheniges (approved):
:van:

PR #593, opened 2020-03-19
Project sort

jjidt:


api/src/controllers/userProjects.js
```diff
@@ -34,11 +35,19 @@ const paramSchema = Joi.compile({
   userId: Joi.number().required()
 })
 
+const querySchema = Joi.compile({
+  page: Joi.number(),
+  perPage: Joi.number(),
```
dheniges: we should set a max number here, maybe 50? Or 100?

api/src/controllers/userProjects.js
```diff
@@ -34,11 +35,19 @@ const paramSchema = Joi.compile({
...
+  sortBy: Joi.string().valid('name', 'updatedAt'),
```
dheniges, line 41: the default sortBy coming from the client is 'createdAt'. Should add it to the list if it's supported, but client should also be updated to default to `updatedAt` since that's what shows in the UI

dheniges (changes requested):
See comments, plus the back/forward/refresh functionality is broken even though the params are in the url. I assume this is because of default state and not reading from the url query params on load

jjidt:
@dheniges Updated, also fixed forward/back for admin users

dheniges (approved):
:realjj:

PR #594, opened 2020-03-19
sync structured value, for now

dheniges:
Hopefully this fixes the problems we've seen. Previous load testing showed this was faster anyway, but it's less scalable


jjidt (approved):
👍

PR #595, opened 2020-03-20
assign random avatar to new users

dheniges:
it would be nice if some of the avatars were more androgynous but 🤷‍♀ 

Maybe Jay's dream of everyone uploading a custom avatar will be realized.

Avatars will only be applied to new users, not existing users


jjidt (changes requested):
Avatar only getting added on multi user add via the admin panel, so needs to be included with the sign-up user as well. Working great for multi user though. Also maybe we could use animal avatars to keep it non-gendered?

dheniges:
@jjidt updated so all user creation flows go through a single coordinator function. Also has new animal avatars now

jjidt (approved):
👍 Nice

PR #596, opened 2020-03-21
socket event listener handling

jjidt:
- refactor for event listener creation/clearing
- sync'ing through hooks


dheniges (approved):
Played around with it a bit, and code seems reasonable. I didn't notice anything wrong happening in the useEffects w/o the docId dependency

PR #599, opened 2020-03-23
increase allowed time for exponential backoff on socketio connection

dheniges:
- it backs off, with exponentials, until it reaches the max attempts


jjidt (approved):
nice 👍wasn't aware that socket io had the built in

PR #601, opened 2020-03-24
UI for Date component

jcontonio:
* Adds Date component with React Datepicker
* Adds custom styles to make it better to work with across devices
* Also removes generic `<button>` styling because that will upset any 3rd party components


dheniges (changes requested):
Good start but needs some tweaks:
- calendar z-index or something off, it doesn't appear above other items and changing months is not possible for me in FF. See screenshots below.
- Creating a new date should probably have it default to today? Not sure if 12am or the current time makes the most sense
- Date field shows cursor when hovering over the text but you cannot edit the text. I think the text should be editable. Currently it defaults to 1969 but if I want it today I have to click 12x[num years]. Also should allow entering any time, I think 15min interval is good for the picker but the field itself doesn't need to be limited to 15min intervals.

Screen Shot 2020-03-24 at 4 48 30 PM
Screen Shot 2020-03-24 at 4 43 24 PM

jcontonio:
* Added: Manual editing of time (test the crap out of this)
* Added: Month and Year select tags
* Added: Styling for weekends, and styling for days that fall outside the month
* Fixed: New dates should be today again
* Fixed: z-index, overflow, and should flip depending on where in the viewport it is

@dheniges ready

dheniges:
This looks very nice. Last thing I've found is the hours parsing  for the valid time message will show hours above 12 which aren't accepted. See screenshot:
Screen Shot 2020-03-25 at 2 30 14 PM

Looks like a quick fix with ` if hours > 12 subtract 12`

jcontonio:
@dheniges all set

dheniges (approved):
This is very nice, definitely one of the nicest if not nicest date/time pickers.

People have never seen a date time picker like this, it's never been done. I inherited a mess of a date time picker, but this has really, really turned it around. It's truly tremendous. You won't believe it when you see it.

PR #602, opened 2020-03-24
mailchimp api

jjidt:
- Add user to mailchimp list if selected when setting password
- If there's no secret key set in env (dev) it will log an ignorable error


dheniges (approved):
Updates looks good. JJ can you add the mailchimp secret to production?

jjidt:
secret added, merging

PR #603, opened 2020-03-25
add env option to turn off logging for web/api

jjidt:


dheniges (approved):
I don't know why I thought this was going to be more complicated but it's perfect

PR #607, opened 2020-03-27
Sync updates

dheniges:
- Merges release, fixes many conflicts
- Updates /users API fetch to use `ids[]=12` format instead of `ids=[12]` which was no longer parsing correctly


jjidt (approved):
tested locally, appears to have fixed it 👍

PR #608, opened 2020-03-27
Reworking image UI and loading / preview

jcontonio:
* All new image UI
* No flashes when updating the image name
* Auto focuses new fields
* Fixes loading / preview states as in it will always show the preview image now, after it starts uploading, resulting in no flashes when updating or adding images


dheniges (dismissed):
Seems like a max-height would be nice on the image so it doesn't take up the whole screen. Portrait dimension images are still very large.

If you edit the name of the image field before uploading an image, it throws the orange "Something went wrong" screen

jcontonio:
Ok, this should be ready

jjidt (approved):
👍

PR #609, opened 2020-03-30
Adjusting position of field users list

jcontonio:
Shouldn't move the content


dheniges (commented):
Can this be absolutely positioned in the center of each field? It's now at the top of the field which makes it harder to tell which field the user icon is in

jcontonio:
Updated to 25% and added a shadow

dheniges (approved):
Ooh this is v. nice

PR #610, opened 2020-03-30
fix mailchimp mane and make it optional secret

dheniges:
this is deployed to dev with feature sync, need to officially roll it in


jjidt (approved):
👍

PR #611, opened 2020-03-30
sync focus update

jjidt:
tested it out, seems like a decent solution, lemme know what you think


dheniges (changes requested):
I think this will go a long way to solving the who's in what field thing, but it does need throttling. Every keystroke fires off an event, maybe debounce it by 1s or so?

jjidt:
half a second seems like the magic number to prevent the focus from ping ponging between fields, updated

dheniges (approved):
👍

PR #613, opened 2020-03-31
Fixing capitalization and projectId bug in delete handler

jcontonio:
The new dashboard had a bug in the delete handler. It was deleting the wrong project id! 😱
This also fixes capitalization throughout the app.


dheniges (changes requested):
I added a project and then went back to the projects list and noticed it hadn't updated (adding a separate bug task for this). So I refreshed, then tried to delete the new project and got this error:
```
Uncaught TypeError: Cannot read property 'name' of undefined
    at eval (NavigatorContainer.js:177)
    at Array.map (<anonymous>)
    at NavigatorContainer (NavigatorContainer.js:175)
    at renderWithHooks (react-dom.development.js:16260)
    at updateFunctionComponent (react-dom.development.js:18347)
    at beginWork$1 (react-dom.development.js:20176)
    at HTMLUnknownElement.callCallback (react-dom.development.js:336)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:385)
    at invokeGuardedCallback (react-dom.development.js:440)
    at beginWork$$1 (react-dom.development.js:25780)
```

jcontonio:
So this deletes the project from the correct page, however... what happens to the amount of projects per page after that I dunno. I guess if I had 5 pages of 10 projects per, removing one would just have 9 on a page until you refreshed?

jcontonio:
This should be ready for review now

dheniges (approved):
Removing that list type works nicely 👍

PR #614, opened 2020-03-31
Simplifying HRs so that we can move the cursor to the end, and see it

jcontonio:
The border and hiding the text just made this a bad experience


dheniges (changes requested):
To confirm, hr's are supposed to look like this now?

Screen Shot 2020-04-01 at 9 02 35 AM

jcontonio:
Yep, for launch anyway. We can't make it work nicely like before and still have the cursor present. I'll put more time in on it after launch

dheniges (approved):
sounds good for now

PR #615, opened 2020-03-31
Sync updates

dheniges:
- DocumentFieldComponent now handles all the sync logic, the way god intended
- Text/String/Number components became dumber (better)
- Fixed a bug with inconsistent dialog popups for remote changes overwriting local stuffs


jjidt (approved):
👍 All looks good to me, let's get it merged and tested on dev

PR #618, opened 2020-04-01
file field

jjidt:
sets up file field, still needs styling


web/src/client/js/components/DocumentFields/DocumentFieldComponent.js
```diff
@@ -28,7 +28,7 @@ const DocumentFieldComponent = ({
     }
   }, [isNewField])
 
-  const dataField = field.type !== FIELD_TYPES.TEXT && field.type !== FIELD_TYPES.IMAGE
+  const dataField = field.type !== FIELD_TYPES.TEXT && field.type !== FIELD_TYPES.IMAGE && field.type !== FIELD_TYPES.FILE
```
dheniges: this should be a whitelist

dheniges (changes requested):
Just one change in the client that I stumbled across the other day, then this looks good to go. I'm ignoring all the copypasta for image field -> file field on the client :)

web/src/client/js/components/DocumentFields/DocumentFieldComponent.js
jcontonio: All set

dheniges (approved):
:+1:

PR #619, opened 2020-04-01
fixing format menu with codemirror

dheniges:
I played around with this but couldn't figure out how we do this without a janky redux-esque connection between the format menu and codemirror instance(s). So it goes through redux. And sync also tracks focus in its context. I'm not sure how we could feed redux data to the sync context. I would like to keep them separate.


jcontonio:
Can confirm format works on this branch 👍

jcontonio (approved):

PR #620, opened 2020-04-01
New: Splitting text fields for new fields

jcontonio:
Updated to do all the API work in one action. Works nicely!

1. Have a big markdown body of text
2. Put your cursor anywhere in it
3. Add a new field

This is going to split the body of the text at your cursor / line position, create your new field, create a new text field with the text after the cursor position, and then reorder the two new fields to put them in the right spot. 

There are 5 unnecessary renders, but that's in `release` too


web/src/client/js/actions/field.js
```diff
@@ -112,6 +115,57 @@ export const copyField = (projectId, currentDocumentId, newDocumentId, field) =>
   }
 }
 
+export const createNewFieldWithTheSplitOfThePreviousFieldAndReOrderThemAppropriately = (
```
jjidt, line 150: 😆 what a name!

jjidt (dismissed):
I like this, makes sense to be able to insert other fields 

A couple things:

- When a new field is created(non split), the cursor no longer goes to that field
- There's a scroll position jump when the new split field is inserted, which can be a tiny bit disorienting

jcontonio:
Update: Switched tracking of lastCreatedFieldId and instead of manually adjusting the CodeMirror instance, calling update endpoint on that field so we get no delay or re-render. Problem is right now calling fetchDocument() doesn't re-render the CodeMirror instance with its new, split value.

Waiting on feature-sync to get in there to see if that'll make things easier. 

Need to call `removeLastCreatedFieldId` from the onFocus handler of `DocumentFieldComponent` once feature-sync is in there.

web/src/client/js/actions/field.js
jcontonio: Solved scroll thing

web/src/client/js/actions/index.js
```diff
@@ -230,8 +231,9 @@ export const Fields = {
   // Blur / Focus events
   blurField: makeActionCreator(ActionTypes.BLUR_FIELD, 'fieldId', 'fieldType', 'fieldData'),
   focusField: makeActionCreator(ActionTypes.FOCUS_FIELD, 'fieldId', 'fieldType', 'fieldData'),
-  // Focus action
-  focusFieldWithId: makeActionCreator(ActionTypes.FOCUS_FIELD_WITH_ID, 'id', 'type'),
+  // Created field Id
+  setLastCreatedFieldId: makeActionCreator(ActionTypes.SET_LAST_CREATED_FIELD_ID, 'fieldId'),
+  removeLastCreatedFieldId: makeActionCreator(ActionTypes.REMOVE_LAST_CREATED_FIELD_ID, 'fieldId'),
```
dheniges: I don't see this action used anywhere?

web/src/client/js/actions/field.js
```diff
@@ -100,6 +103,18 @@ export const focusField = (fieldId, fieldType, fieldData) => {
   }
 }
 
+export const setLastCreatedFieldId = (fieldId) => {
+  return async (dispatch) => {
+    dispatch(Fields.setLastCreatedFieldId(fieldId))
+  }
+}
+
+export const removeLastCreatedFieldId = (fieldId) => {
```
dheniges: Same thing, I don't see this called anywhere

web/src/client/js/components/DocumentFields/DocumentFieldsContainer.js
```diff
@@ -55,6 +55,29 @@ const DocumentFieldsContainer = ({
...
+      // It should append a new text field to the end of the document, making it seem as though the
+      // user is clicking to continue the document body
+      const newName = getNewNameInSequence(fields, FIELD_TYPES.TEXT)
+      createField(projectId, documentId, FIELD_TYPES.TEXT, {
+        name: newName,
+        type: FIELD_TYPES.TEXT
+      })
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
dheniges: this should at least have `documentReadOnlyMode` as a dependency, and fields, since those can change. Otherwise this creates a reference to fields on the first render and the "getNewNameInSequence" will keep returning the same name.

dheniges (changes requested):
A few code things, but the actual splitting itself seems very nice

web/src/client/js/actions/index.js
jcontonio: There was a place we were going to call this... @jjidt do you remember? I know we were waiting until after sync got in.

jjidt: I really can't remember what the deal was with that, where is that last created field id state getting updated now? seems like it's working how we want

dheniges (approved):
woooo

PR #621, opened 2020-04-02
Fixing multiple renders of component

jcontonio:
DocumentFieldsComponent was re-rendering a lot due to its `fields` prop changing a lot on an update. To reproduce this, do the following on `release`:

1. Add the following to `DocumentFieldsContainer`
  ```
  useEffect(() => {
    console.log('render container')
  }, [fields])
 ```

2. Add the following to `DocumentFieldsComponent`
  ```
  useEffect(() => {
    console.log('render component')
  }, [fields])
 ```

3. Create a new document with 1 text field
4. Add the letter `a` to the text field, and see 2 renders, then the PUT, then a bunch more renders

On this PR, the same result will be 1 render from the container, 1 render from the component, after the PUT.

Before and after screenshots:

Screen Shot 2020-04-02 at 11 54 29 AM
Screen Shot 2020-04-02 at 11 49 31 AM

### DIFF
--- a/web/src/client/js/components/Document/DocumentComponent.js
+++ b/web/src/client/js/components/Document/DocumentComponent.js
@@ -29,8 +29,6 @@ const DocumentComponent = ({
   const titleRef = useRef()
   const { data: userProjectAssignments = {}, loading: assignmentLoading } = useDataService(dataMapper.users.currentUserProjectAssignments())
 
-  console.log('Render document')
-
   // If we've exited fullscreen, but the UI is still in fullscreen
   // This happens when hitting the escape key when in fullscreen
   useEffect(() => {
@@ -46,6 +44,9 @@ const DocumentComponent = ({
       window.document.removeEventListener('webkitfullscreenchange', fullScreenChangeHandler, false)
     }
   })
+
+  if (!document.id) return null
+  console.log('Render Document Component')
 
   const docKey = document ? document.id : 0
   const projectAssignment = userProjectAssignments[Number(projectId)]
--- a/web/src/client/js/components/Document/DocumentComponent.js
+++ b/web/src/client/js/components/Document/DocumentComponent.js
@@ -29,6 +29,8 @@ const DocumentComponent = ({
   const titleRef = useRef()
   const { data: userProjectAssignments = {}, loading: assignmentLoading } = useDataService(dataMapper.users.currentUserProjectAssignments())
 
+  console.log('Render document')
+
   // If we've exited fullscreen, but the UI is still in fullscreen
   // This happens when hitting the escape key when in fullscreen
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


dheniges:
I added logging in the container, the fields, and field document components and found this saves 1 call to the document component, buuuut, it introduces a depth warning. I added logging at the top of each component to see every time they were invoked.

Release:
Screen Shot 2020-04-02 at 1 03 02 PM

This branch:
Screen Shot 2020-04-02 at 1 03 02 PM

Full warning:
Screen Shot 2020-04-02 at 12 59 54 PM

The concerning thing to me is the container gets called _a lot_. As in, 20 or 30 times, which seems like something higher up the chain needs fixing.

I'm not sure if part of the issue is the useEffect here, as that's run _after_ render, I think `useMemo` may be more appropriate here and might fix the warning?

dheniges:
I added logging in the container, the fields, and field document components and found this saves 1 call to the document component, buuuut, it introduces a depth warning. I added logging at the top of each component to see every time they were invoked.

Release:
Screen Shot 2020-04-02 at 1 03 02 PM

This branch:
Screen Shot 2020-04-02 at 1 03 02 PM

Full warning:
Screen Shot 2020-04-02 at 12 59 54 PM

The concerning thing to me is the container gets called _a lot_. As in, 20 or 30 times, which seems like something higher up the chain needs fixing.

I'm not sure if part of the issue is the useEffect here, as that's run _after_ render, I think `useMemo` may be more appropriate here and might fix the warning?

dheniges:
I added logging in the container, the fields, and field document components and found this saves 1 call to the document component, buuuut, it introduces a depth warning. I added logging at the top of each component to see every time they were invoked.

Release:
Screen Shot 2020-04-02 at 1 03 02 PM

This branch:
Screen Shot 2020-04-02 at 1 03 02 PM

Full warning:
Screen Shot 2020-04-02 at 12 59 54 PM

The concerning thing to me is the container gets called _a lot_. As in, 20 or 30 times, which seems like something higher up the chain needs fixing.

I'm not sure if part of the issue is the useEffect here, as that's run _after_ render, I think `useMemo` may be more appropriate here and might fix the warning?

dheniges (approved):
I took the liberty of removing the logging since I was logging stuff out there anyway.

So moving the fields logic into useEffect or useMemo doesn't change the number of renders. Where render = DOM update, measured by `useEffect(() => { ...post-render code...})` with no dependencies. But, not doing that field sort every time it's called should make it faster, the sorting code is invoked significantly less.

In terms of actual renders (again, not component function invocations, but actual renders), the addition of returning null if the document has no id saves two renders when initially loading a document.

In the future we may want to explore ways to reduce the number of update renders, which stands at 5 currently.

:+1:

PR #622, opened 2020-04-02
project delete/create loading fix

jjidt:
- gets rid of list loading state references
- fixes create and delete project reducers

PR'd into jay's correct case PR, there was some overlap, this would need merged before that one goes into release


jcontonio (approved):
Works nicely

PR #623, opened 2020-04-03
Web - Updating packages for audit

jcontonio:
Should fix the security audit on web


dheniges (commented):
Screen Shot 2020-04-03 at 10 46 19 AM

I can't get web to run -- does it work for you after doing a `make installweb`?

jcontonio:
https://github.com/npm/npm/issues/3497

So it seems to fix the preinstall script we'd have to run npm install with the `--unsafe-perm` flag. I am wondering if this is causing dotenv and other packages to not be installed? I'm not sure why this would work locally but not in Docker other than that?

PR #624, opened 2020-04-03
Feature – file field

jcontonio:


jcontonio:
Seems like this is ready to go

dheniges (changes requested):
I think the last thing this needs is copying the file asset on S3 on publish, like we do with images

dheniges (approved):
:peepee:

PR #626, opened 2020-04-07
SyncMaster 2000

dheniges:
- Fixes sync
- Rewrites the dependencies of FieldTextComponent so when values change the appropriate CodeMirror events and callbacks get updated
- Removes the new fieldFocus emitter for sync because of the seven fresh hells it unleashes. See code comment
- Also fixes sync


jjidt (approved):
👍 Doing gods work here

PR #627, opened 2020-04-08
single user sync edit field support

dheniges:
PR'ing this now, but date and image fields need to be updated to set/unset focus still.

This will work on the date field's name currently.


jjidt (approved):
👍

jjidt:
LMMTFY

PR #628, opened 2020-04-08
File field styling

jcontonio:
Screen Shot 2020-04-08 at 12 24 49 PM

* Styles File field
* Moves File field to data menu
* Updates Format menu and Content menu so they're navigable via arrow keys


jcontonio:
Forgot to hook up file extension check for the blacklist. Just updated

jjidt (approved):
👍 very nice

PR #629, opened 2020-04-08
merge release into sync

jjidt:
Doing this as a PR rather than direct merge, it did some goofy things with replacing code it shouldn't have. I think I caught it all, but worth another look


dheniges (approved):
I tested out a bunch of sync stuff and all looks well :+1:

PR #630, opened 2020-04-09
date focus/lockout

jjidt:
- handle focus/blur events for dates
- lock out non-focused users from date and time selection

PR'd into the release -> sync PR, need to change base to feature-sync once that is merged


dheniges (changes requested):
It's better, but here's what I'm still seeing:
1. The date field doesn't update automatically for other users. User A and B are in the same doc. User A is the only one editing the date field, they change the time/date/field name. User B will re-fetch the document, but the new values don't show up until User B refreshes.
2. The time field is manually editable, but it does not set the focus/blur events

After those issues are updated, I think we pass to Jay to do the following:
- Update readOnly styling to show the field is disabled. Right now you still see the cursor on hover and nothing looks disabled, so it's confusing when you click on the calendar icon why the calendar doesn't open
- Expand the date select calendar to open when clicking both the icon and the date string itself.

jjidt:
@jcontonio This is ready for you if you want to make the changes Dirk proposed

jcontonio:
@jjidt updated to have the date open the calendar on click

dheniges (commented):
Everything looks good except that time field is still not locking out for me. Am I missing something there?

jjidt:
updated, just missing the readOnly flag on that specific input

dheniges (approved):
Nice work.

PR #632, opened 2020-04-14
image remote user lockout

jjidt:
- locks out other users if someone is making an image name change or in settings mode
- selecting an initial file doesn't focus/blur unfortunately (events not supported)
- file component is a little different, since the name is handled in the parent component, it will do the remote change acceptance dialogue, but the settings lock out in the same way


dheniges (changes requested):
The image component works nicely, but the file component conflict dialog doesn't work for the file field name. Choosing "discard" doesn't show the latest remote changes until you refresh. I haven't looked at the code, but can this not function like the date and other fields that lock it out?

jjidt:
updated and tested with the name change lockout for file fields

dheniges (approved):
👍 huzzah

PR #634, opened 2020-04-15
date publish validation fix

jjidt:
date values coming back from the db are Date instances, need to convert these to strings to validate during publish


dheniges (approved):
👍

PR #637, opened 2020-04-16
Updates: Dims out fields that are being edited by someone else

jcontonio:
Grays them out and disables any mouse interaction

### DIFF
--- a/web/src/client/js/components/DocumentFields/DocumentFieldComponent.js
+++ b/web/src/client/js/components/DocumentFields/DocumentFieldComponent.js
@@ -22,6 +22,12 @@ const DATA_FIELD_TYPES = [
   FIELD_TYPES.STRING,
 ]
 
+const FIELDS_THAT_DISABLE_WHEN_EDITING = [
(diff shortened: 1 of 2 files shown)


dheniges (changes requested):
This disables all field types, including text fields. I don't think we want that?

jcontonio:
Oh duh

jcontonio:
Updated

web/src/client/js/components/DocumentFields/DocumentFieldComponent.js
```diff
@@ -22,6 +22,12 @@ const DATA_FIELD_TYPES = [
   FIELD_TYPES.STRING,
 ]
 
+const FIELDS_THAT_DISABLE_WHEN_EDITING = [
```
dheniges: don't need this now?

dheniges (changes requested):
Works good now, looks like just a leftover constant whitelist you started?

dheniges (approved):
:sealeel:

PR #640, opened 2020-04-17
Fixes: Arrow key navigation of menus in production

jcontonio:
MenuItem's displayName was getting mangled in production. Setting an explicit displayName on the component fixed an issue where you couldn't arrow nav menus in prod.

Also fixes running prod locally


dheniges (approved):
👍

PR #641, opened 2020-04-17
Updates: scroll position when splitting a text field

jcontonio:
No longer jumps things around. We save the scroll position of document__fields before re-render, and set it back to that. 

Adding a new field shouldn't move the document at all now, feels much nicer.


jjidt (approved):
👍 Works great, makes the split feel seamless

PR #642, opened 2020-04-17
Fixes: File field's name widths were off when adding a new file

jcontonio:
Try creating a file with the name `Screen Shot 2020-04-16 at 11.19.50 AM.png` on release and you'll see:

image

Now you should see the normal field name width like everything else


dheniges (approved):
:+1:

PR #643, opened 2020-04-17
Updates: Popup dialog when discarding changes has an arrow now

jcontonio:


dheniges (changes requested):
The arrow placement seems a little odd, is there an option to center it or does it just go where popper puts it?

More importantly, I noticed the conflict dialog always appears at the top of the field, so if you're typing in a longer text field you won't see it pop up. Can it float like the user icons do?

Screen Shot 2020-04-20 at 11 10 30 AM

dheniges (approved):
beautiful

PR #644, opened 2020-04-17
New: Notify users when there's a new version of the web app

jcontonio:
Prompts user to refresh when app bundle is different.
You can only test this in prod, when the app bundle has changed from the cached one. 

Looks like this:

Screen Shot 2020-04-17 at 1 05 36 PM


jjidt (commented):
I like it, maybe use `changes` instead of updates in second sentence, whaddya think?

jjidt (approved):
👍

PR #645, opened 2020-04-17
Fixes: Prevents useBlur hook from triggering when using mouse

jcontonio:
For the Content and Format menus


dheniges (changes requested):
The format menu doesn't close after you make a selection now, it stays open. This is also true on mobile for some reason. In release the menu closes after making a selection.

dheniges (approved):
👍  nice this seems to work well across everything

PR #646, opened 2020-04-18
field name sync fix

jjidt:
I swear we had some of this working, maybe a bad merge
- does name change checks and updates via dialogue
- upates in a useEffect


jjidt:
updated to handle name change updates in the outline

dheniges (approved):
👏  solves all my problems

PR #647, opened 2020-04-22
Fixes: doc toolbar from appearing when it shouldn't

jcontonio:
This fixes the bug, but I'm not sure why `currentResource` still returns a document when `matchResults` is null.


jjidt (approved):
👍 Fixes the bug, useDataService fix coming through next

PR #648, opened 2020-04-22
Sync field Popper name-disappearing bugfix / Reduxify sync

jjidt:
- removes of socket provider context flow
- redux all the user sync actions/ reducers
- cleans up field change logic in DocumentFieldComponent
- adds additional loading check in DocumentContainer to reduce render count


web/src/client/js/components/DocumentUsers/DocumentUsersContainer.js
```diff
@@ -1,23 +1,29 @@
...
-const DocumentUsersContainer = ({ fetchUsersWithIds, usersById, usersLoadedById }) => {
+const DocumentUsersContainer = ({ fetchUsersWithIds, usersById, usersLoadedById, activeDocumentUsers, emitJoinDocumentRoom, emitLeaveDocumentRoom }) => {
   const { documentId } = useParams()
 
-  useSyncDocumentRoom(documentId)
-
-  const { state: socketState } = useDocumentSocket()
+  // User Sync for document room join/leave
+  useEffect(() => {
+    if (document) {
```
dheniges:
```suggestion
    if (documentId) {
```

I don't see where document is defined here, is this supposed to be documentId?

web/src/client/js/components/DocumentUsers/DocumentUsersContainer.js
```diff
@@ -1,23 +1,29 @@
...
+      emitJoinDocumentRoom(documentId)
+    }
+    return () => {
+      emitLeaveDocumentRoom()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dheniges: if the above change of document -> documentId is made this comment is no longer needed

dheniges (changes requested):
I think there are some things left to iron out in these updates. Namely, joining and leaving rooms is inconsistent. It looks like the leaveRoom event is emitted without an id most of the time, and other times the `userChange` event is emitted with only a single user id that removes the other person that was in the document. I'm not sure what the userChange vs. join/leave room events differentiate?

dheniges:
Also like your proposed name change for `userChange` to become `documentRoomUsersUpdated`

dheniges (approved):
:realjj:

PR #651, opened 2020-04-24
useDataService bugfix

jjidt:
- this fixes situations where a component is going
from:  having loaded data available through useDataService
to: a condition where we want it to not fetch a resource and return null automatically

- returns `false` in returnNull `getLoadingStatusFromState` functions, which signals to getDataService that it doesn't need to call the fetch function at all, and to set the exposed state to the new `null` value


dheniges (approved):
:tada:

PR #654, opened 2020-04-24
Changes: Tracking state for format field in UI state

jcontonio:
Fixes the format menu focus
Fixes the split field functionality


dheniges (approved):
👍

PR #655, opened 2020-04-24
fix field name flash

jjidt:
- adds a nifty usePrevious hook that will return the previous value on each re-render
- adds a check for generic name changes and image name changes, if the incoming field.name is the same as the previous one, we know not to update the current.value

### DIFF
--- a/web/src/client/js/components/DocumentFields/DocumentFieldComponent.js
+++ b/web/src/client/js/components/DocumentFields/DocumentFieldComponent.js
@@ -5,6 +5,7 @@ import { connect } from 'react-redux'
 
 import { FIELD_TYPES, SYNC_SINGLE_USER_EDIT_FIELDS } from 'Shared/constants'
 import { currentUserId } from 'Libs/currentIds'
+import usePrevious from 'Hooks/usePrevious'
 
 import DocumentUsersComponent from 'Components/DocumentUsers/DocumentUsersComponent'
 import { Popper } from 'Components/Popper/Popper'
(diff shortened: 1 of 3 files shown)


dheniges (changes requested):
I'm still getting the flash on this, it seems like there's a point where the previousValue matches the field name in redux here after a render but before the PUT response is received

jjidt:
added an isUpdating check there as well and it seems to behave better when wildly clicking around and updating things. This probably won't completely solve it for super slow connections, so we might want to implement some kind of component update date vs. lastUpdatedAt date check in the future, possibly?

dheniges (commented):
I think the isUpdating check fixes the issue. I removed the previousField code and it still works as expected, so I don't think we want/need that previousField check here?

web/src/client/js/components/DocumentFields/DocumentFieldComponent.js
```diff
@@ -83,12 +85,12 @@ const DocumentFieldComponent = ({
...
-    // Accept remote changes to field if it's not focused and the conflict popper isn't open
-    if (!isCurrentlyFocusedField && !showConflictPopper) {
+    // Accept remote changes to field if it's not focused, the conflict popper isn't open, and the field value has changed
+    if (!isCurrentlyFocusedField && !showConflictPopper && previousField !== field && !isUpdating) {
       setCurrentValue(field.value)
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [field.id, field.value])
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [field, isUpdating])
```
dheniges: I think this needs previousField as a dependency

dheniges (changes requested):
This fixes my flashing problems. I think it needs one dependency update in the useEffect though

jjidt:
updated

dheniges (approved):
:realjj:

PR #656, opened 2020-04-24
Changes: Discard changes UI update

jcontonio:
* Removed the popper and used a tiny popover style
* Fixes the height problem of fat fields


dheniges (approved):
👍

PR #658, opened 2020-04-28
Updating CodeMirror and removing dividers

jcontonio:
* Updates to 5.53.2 for a CodeMirror.refresh() bug with wrapping text
* Removes "Content" and "Data" from content menu


dheniges (approved):
👍

PR #660, opened 2020-04-28
Fixes: File field margin & field name length overflow

jcontonio:
Screen Shot 2020-04-28 at 10 29 13 AM


web/src/client/js/components/DocumentFields/DocumentFieldComponent.js
```diff
@@ -249,13 +236,11 @@ const DocumentFieldComponent = ({
               <span className="document-field__name-label">{fieldLabels[field.type]}</span>
               <input
                 defaultValue={field.name}
-                maxLength={fieldNameMaxLength}
```
dheniges, line 252: I think this still needs max length, otherwise I can put in a name that's too long and get an API error that doesn't show up in the UI

dheniges (changes requested):
I think maxLength still needs to be on the input?

dheniges (approved):
👍

PR #661, opened 2020-04-28
Fix project filtering

jjidt:
Always pass projects in array to maintain sort order

### DIFF
--- a/web/src/client/js/components/Dashboard/DashboardComponent.js
+++ b/web/src/client/js/components/Dashboard/DashboardComponent.js
@@ -33,7 +33,7 @@ const DashboardComponent = ({
   const history = useHistory()
   const hasProjects = projects.length > 0
   const { myProjects, notMyProjects } = projects.reduce((cur, project) => {
-    if (projects.createdBy === currentUserId) {
+    if (project.createdBy === currentUserId) {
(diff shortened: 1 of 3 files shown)


web/src/client/js/components/Dashboard/DashboardComponent.js
```diff
@@ -31,17 +31,15 @@ const DashboardComponent = ({
...
   const [showMyProjects, toggleMyProjects] = useState(false)
   const history = useHistory()
-  const objectKeys = Object.keys(projects)
-  const hasProjects = objectKeys.length > 0
-  const { myProjects, notMyProjects } = objectKeys.reduce((cur, key) => {
-    if (projects[key].createdBy === currentUserId) {
-      cur.myProjects[key] = projects[key]
+  const hasProjects = projects.length > 0
+  const { myProjects, notMyProjects } = projects.reduce((cur, project) => {
+    if (projects.createdBy === currentUserId) {
```
dheniges:
```suggestion
    if (project.createdBy === currentUserId) {
```

Drop the `s` on `projects` or this doesn't work

dheniges (changes requested):
One small tweak for codekind

jjidt:
updated

dheniges (approved):
🦖

PR #663, opened 2020-04-28
Fixes: Public project checkbox shows the correct value

jcontonio:


jjidt (approved):
👍 tested, works great

PR #666, opened 2020-04-29
fix empty payload validation

dheniges:
I only saw this on multer-backed endpoints for avatars, not really sure how this ever worked, but maybe broke with the joi upgrade?


jjidt (changes requested):
We don't want to call Joi.compile inside a function, probably want to pre-compile that in module scope

dheniges:
@jjidt added a sweet new schema, good call

jjidt (approved):
👍 Works great, no errors

PR #667, opened 2020-04-29
Fixes: focusing of new fields

jcontonio:
* Should scroll new fields into view, no matter the field type
* Should focus the string, number, or text field
* Should work on both split fields and not split fields


jjidt (approved):
Tested with all the field types, works great

PR #673, opened 2020-04-30
bump org rate limit

dheniges:
Puts it at 1000 per org, so for a 5 user org each user gets 200 req/min which is hard to hit, I have to refresh the app ~18 times in a row to hit 200 req/min. Each refresh is 11 requests


jjidt (approved):
:shockandawe:

jjidt:
LMMTFY

PR #676, opened 2020-05-01
update api packages for securitayz

dheniges:
- Major jest version upgrade
- Rest are minor/patch version fixes


jjidt (approved):
👍 still works

PR #677, opened 2020-05-01
Fixes: Hiding filters for read-only users

jcontonio:
If a user isn't an admin, or if everyone cannot create projects: 

image


dheniges (approved):
👍

PR #678, opened 2020-05-01
emit join document room on focus

jjidt:
only happens if there's not a currently set document room in state
couldn't repro locally, so we'll need to test on dev


dheniges:
We'll go ahead and close. For historical purposes, the currentDocumentRoomId used here in redux state is only set by actions the user takes in the web app, it is never updated from incoming sync events, so this never got triggered. The underlying bug that drove this PR was fixed by a bugfix on the sync server wanting a string and getting a number id.

PR #681, opened 2020-05-07
Adding terms of service checkbox

jcontonio:
Adds checkbox to sign-up entrypoint


dheniges (changes requested):
Probably worth adding the terms url to the constants so it's in one spot

jcontonio:
Updated

dheniges (approved):

dheniges (approved):

PR #682, opened 2020-05-07
cast to string for documentId in sync focusField event

dheniges:
- Fixes bug where focusField wasn't forcing user to join room if user was not in room (in this case caused by redis expiring the user in the room)
- Adds logging to help identify this issue going forward

I had many theories about what was happening, and most of them were wrong. At least I validated a lot of the sync functionality though?


jjidt (approved):
👍 nice

PR #683, opened 2020-05-07
Trial Emails

jjidt:
- Send emails on trial_will_end event
- Check the org on invoice.payment_failed and send a trial ended email if applicable

The invoice.payment_failed event should also trigger the `await billingCoordinator.fetchCustomerAndSetSubscriptionDataOnOrg(org.id)` function, which will update the cached subscriptionStatus on the org to PAST_DUE

best way to test this is to: 
- get it up on dev
- create a new org
- in the stripe panel set the trial end date to 4 days from now
- sit back and wait for some emails


dheniges (approved):
The code looks good, I'm not cool enough to ngrok / my computer will go to sleep so I'm down on testing this on dev. I think we'll want to add some more info to the emails but we can do that separately

PR #685, opened 2020-05-11
handle trial end status/email

jjidt:


dheniges (approved):
I think this should handle the scenarios. Which probably means there's a weird scenario I'm not thinking of

PR #687, opened 2020-05-12
add missing constant

jjidt:


dheniges (approved):
bold moves JJ, bold moves

PR #691, opened 2020-05-13
log subscription status changes on org

jjidt:
- bonus optimization here to only update status on org if it actually changes
- doesn't really fall into audit log realm, so just using logger here

### DIFF
--- a/api/src/coordinators/billing.js
+++ b/api/src/coordinators/billing.js
@@ -8,6 +8,8 @@ import { BILLING_PUBLIC_FIELDS } from '../constants/billingPublicFields'
 import { PLANS, MULTI_USER_DEFAULT_SEAT_COUNT, STRIPE_STATUS, ORG_SUBSCRIPTION_STATUS } from '../constants/plans'
 import { getOrgSubscriptionStatusFromStripeCustomer } from '../libs/orgSubscription'
 import slackIntegrator from '../integrators/slack'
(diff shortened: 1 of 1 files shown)


api/src/coordinators/billing.js
```diff
@@ -249,7 +251,13 @@ const fetchCustomerAndSetSubscriptionDataOnOrg = async function(orgId) {
   const customer = await stripeIntegrator.getCustomer(org.stripeId)
   const subscriptionStatus = getOrgSubscriptionStatusFromStripeCustomer(customer)
 
-  const orgUpdateData = { subscriptionStatus }
+  const orgUpdateData = {}
+
+  if (subscriptionStatus !== org.subscriptionStatus) {
+    orgUpdateData.subscriptionStatus = subscriptionStatus
+    // log subscription changes
+    logger(LOG_LEVELS.INFO, { type: 'org subscription status update', message: `The subscription status for organization ${org.name} has changed from ${org.subscriptionStatus} to ${subscriptionStatus}` })
```
dheniges: Let's add the org id here, since the org name is changeable. `status for organization ${name} (${id})` or something

dheniges (changes requested):
Tweak on the logged data

jjidt:
updated

dheniges (approved):
👍

PR #694, opened 2020-05-18
Adjusting locked statuses

jcontonio:
If a user is in TRIAL_ENDED, they are locked, but they should get different messages than if they have one of the other locked subscription statuses

### DIFF
--- a/web/src/client/js/components/Admin/AdminNotices/AdminNoticesComponent.js
+++ b/web/src/client/js/components/Admin/AdminNotices/AdminNoticesComponent.js
@@ -8,7 +8,6 @@ import {
   PLAN_TITLES,
   PLAN_TYPES,
   SUPPORT_EMAIL,
-  TRIAL_ENDED,
   TRIALING_STATUSES
 } from 'Shared/constants'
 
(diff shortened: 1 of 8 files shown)


dheniges (commented):
Works, but I'm not a big fan of multiple logic gates in a render function. Instead of checking multiple user roles, or checking the org subscription status inline it reads much nicer, and is less error-prone, to move logic into functions. ie `{getAdminNotice()}` that can only return one notice, vs the hodgepodge of { logicCheck && .... } in the render which could render a bunch of things accidentally and it's hard to test for. Verse a function with a switch statement that can only return one thing.

Your call if you want to update now though, this works currently.

jcontonio:
Are you suggesting something along the lines of this? Because I think it's way less readable. Each case has slightly different markup, so returning an object with JSX in it isn't really any better.

I vote either adding comments above each conditional, or move each notice into its own component so that it's less stuff in the file and easier to read. 

Unless you're saying suggesting something else?

```
function getAdminNotice() {
    if (TRIALING_STATUSES.includes(organization.subscriptionStatus)) {
      return {
        // return markup
      }
    }
    if (lockedAccountStatusesMinusInactive.includes(organization.subscriptionStatus)) {
      return {
        // return markup
      }
    }
    switch (organization.subscriptionStatus) {
      case ORG_SUBSCRIPTION_STATUS.TRIAL_ENDED: {
        // return markup
      }
      case ORG_SUBSCRIPTION_STATUS.INACTIVE: {
        // return markup
      }
      case ORG_SUBSCRIPTION_STATUS.INACTIVE: {
        // return markup
      }
      case ORG_SUBSCRIPTION_STATUS.PENDING_CANCEL: {
        // return markup
      }
    }
  }
```

dheniges:
My concern is that you can only do && operators and ternaries in JSX, so each condition can pass/fail and every condition is evaluated. It would be easy to introduce a bug that caused multiple notices to show up that shouldn't, for instance if we tweak how subscriptions are handled or add/change an org status. If the logic is in a function you can guarantee it only matches one condition by returning each markup notice as soon as a condition passes. And for readability you can have a logic function that returns separate markup functions if you want. 

TL;DR: My concern is the current logic is the equivalent of a switch() statement with no breaks or returns.

dheniges (approved):
much cleaner/more better 👍

PR #695, opened 2020-05-20
new admin route for updating org billing status and fetching orgs

dheniges:
Required by https://github.com/BonkeyBong/portway-ops/pull/3

The fetchAll on BusinessOrgs should be paginated at some point, but running the job locally with 70 orgs is fine. When we get more orgs we'll probably need a different runner structure to handle these updates


jjidt (approved):
👍 Ran through the flow with Postman and successfully updated an out-of-sync org

PR #696, opened 2020-05-20
Changes: Removes user menu border when avatar present

jcontonio:
It looked weird when an avatar wasn't filling the circle


dheniges (dismissed):
This doesn't run, app errors out in the browser.

With the default avatars now isn't it impossible for new users to not have an avatar? I dunno if that simplifies things by not trying to handle both cases nicely

jcontonio:
Fixed, was crashing on refresh. Sorry!

jjidt (approved):
👍Looks good :dumps:

PR #699, opened 2020-05-20
Fixes: Image fields with tiny images displayed funky

jcontonio:
Try an image that's like 24x24 and you'll see that it no longer looks terrible


jjidt (approved):
👍 looks good

PR #701, opened 2020-05-27
don't send email when trial user's account canceled by stripe

jjidt:


api/src/coordinators/stripeEvent.js
```diff
@@ -28,6 +28,12 @@ async function handleEvent(event) {
       break
     }
     case 'customer.subscription.deleted': {
+      // if customer never had billing info, assume they're transitioning from a trial PAST_DUE status
```
dheniges, line 31: could this get explicitly checked with previousAttributes.status === PAST_DUE ?

dheniges (approved):
why not, worst case scenario someone with a canceled account gets an email

PR #711, opened 2020-06-10
API image to stretch slim

dheniges:
- Updates to shell scripts to work with stretch-slim node image variant
- Overall image size dropped from ~150mb to 70mb
- Some debug output in the shell scripts


jjidt (approved):
👍 woooo!

PR #712, opened 2020-06-10
Feature image upgrades

dheniges:
Latest node version upgrades and move to debian slim images


dheniges:
Need to nail down the issue with tests. Seems like the composition in codefresh doesn't have a migrated database, so maybe the migrate is running before the db service is up?

jjidt (approved):
🎉

PR #713, opened 2020-06-10
Docker web updates

dheniges:
- Move base image to stretch-slim
- Works with rapid7 logging
- Upgrades to node 12.18.0
- Reduces image size from 250ish mb to 61mb thanks to multi-stage docker build


jjidt (approved):
👍

PR #715, opened 2020-06-12
Fix tests with new docker images

dheniges:
For whatever reason, the database needed a few more seconds to startup. The migrate command was running before the db was ready for commands, even though it was responding on the host and port.

At least, that seems to be the case.


jjidt (approved):
👍 nice

PR #720, opened 2020-06-22
Fixes: new document item was hidden on scrolled lists

jcontonio:
If you had a long document list, clicking "New document" or Ctrl-N wouldn't scroll the list up to the top


dheniges (approved):
:+1: v nice

PR #722, opened 2020-06-23
Fixes: Inserting fields at the beginning of a text field

jcontonio:
* Fixes the bug where if your cursor was at 0,0 in a text field, the text field would be duplicated 
* Updates the "split" so that if your cursor is at 0,0 in a text field, the new added field will go before the cursor


dheniges (changes requested):
It works nicely if the cursor's at the beginning of the text field, but when the cursor's anywhere in the middle it's not splitting the field for me, it's just adding the new inserted field at the bottom of the doc

dheniges (approved):
I'm going to blame webpack. Fresh start and hard refresh and it's working nicely

PR #725, opened 2020-06-30
Document slug field

dheniges:
Adds a new `slug` field to Document which gets auto-generated on document create.

We probably want a job to create and set slugs on existing documents.

### DIFF
--- a/api/src/constants/resourcePublicFields.js
+++ b/api/src/constants/resourcePublicFields.js
@@ -2,11 +2,22 @@
 import resourceTypes from './resourceTypes'
 import globalPublicFields from './globalPublicFields'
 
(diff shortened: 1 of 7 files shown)


jcontonio (approved):
Works tried with emoji in title name as well 👍

Should we have a script that goes through and manually does this to all existing documents? Since we're live now, I'm not sure it's great to have a bunch of documents without slugs?

dheniges:
> Works tried with emoji in title name as well 👍
> 
> Should we have a script that goes through and manually does this to all existing documents? Since we're live now, I'm not sure it's great to have a bunch of documents without slugs?

you didn't read my PR description 😿

jcontonio:
I definitely didn’t!

PR #726, opened 2020-07-01
documentVersion slug + db job k8s config

dheniges:
k8s config for database jobs, used in conjunction with image from https://github.com/BonkeyBong/portway-ops/pull/4


jjidt (approved):
👍

PR #727, opened 2020-07-01
Document slug publishing

dheniges:
Handles slug publishing with documentVersions correctly.

I thought about unique constraints for these, but I dunno that we want to enforce that.


jjidt (approved):
👍

PR #728, opened 2020-07-07
project export coordinator

jjidt:
write md files, download assets, and zip it up

TODO:

- .env temp file path
- controller perms
- worker/upload to s3


api/src/integrators/zip.js
```diff
@@ -0,0 +1,73 @@
...
+  })
+}
+
+const getRecursiveDirectoryFiles = async function(dir) {
+  const entries = await fs.promises.readdir(dir, { withFileTypes: true })
+  const files = await Promise.all(entries.map((entry) => {
+    const entryPath = path.resolve(dir, entry.name)
+    return entry.isDirectory() ? getRecursiveDirectoryFiles(entryPath) : entryPath
+  }))
+  return Array.prototype.concat(...files)
```
dheniges, line 68: This is returning a new array of the files, which are arrays or strings it looks like. But why the new array vs returning `files` here?

dheniges (changes requested):
One code question (see comment), and then:
- if I have a date field, with or without a value, I get an error: cannot read property `toISOString` of null
- A project with 1 document with a single text field just hangs when I try to export it.

I'm running the export through postman

api/src/integrators/zip.js
jjidt: just a shortcut to flatten a multi dimensional array, which files ends up being

dheniges:
This work is now in the feature branch via another PR, closing

PR #734, opened 2020-07-16
Fixes: Document list mobile view

jcontonio:
On release, if you have a short document list (1-2 docs) and you clicked the `...` menu, it'd be cut off. The list wasn't expanding all the way down the view. Now it does.


dheniges (approved):
Looks good

PR #738, opened 2020-07-20
Export party

dheniges:
Functional worker implementation with docker compose for generating export zips.

- Moves all the things to the worker
- Implements BeeQueue for workers, coordinated with redis pub/sub
- Fixes bugs and issues

## TODOs
- [x]  Add S3 upload after zip is created
- [x]  Move and update tests from api to worker
- [ ]  Kubernetes configuration
- [x]  Add logging to rapid7


jjidt:
@jcontonio This is working if you wanna check it out, we can get it merged into the feature branch and then start on the client

The rest of the items I can do in separate PRs

jcontonio (approved):

PR #739, opened 2020-07-20
Adds editable document slug

jcontonio:
Adds a "Document info" panel to the Document outline

~This still needs "slug" turned on in the API~


dheniges (changes requested):
A few items from testing:

At shorter screen heights the display gets a bit messed up:
Screen Shot 2020-07-24 at 2 42 17 PM

The UI validation is a bit confusing, it just turns red if you put in an invalid character. What if we just add a note "only the letters a-z and numbers 0-9 are accepted."

The "slugify" utility function in the client looks like it's not used?

dheniges (commented):
Works purdy nice 👍 

However, there's that unused slugify utility added. Drop that?

dheniges (approved):

PR #744, opened 2020-07-31
worker logging

jjidt:
- definitely need to make this a reusable npm lib at some point, we have the same thing happening in 4 different places
- still needs a worker logger namespace set up on insight and the token added to 1pass


dheniges (changes requested):
The constants aren't quite right here, there's no index in the constants root, and there's an empty logLevels constants file

jjidt:
updated

dheniges (approved):
Sweet, works on r7

PR #747, opened 2020-08-04
Updating export ui

jcontonio:
Small updates, mostly language and different button states to make it obvious what's happening


jjidt (changes requested):
Did ZipIcon get commited? I'm getting an error that looks related to that

jcontonio:
Updated

jjidt (approved):
👍 looks good

PR #748, opened 2020-08-04
Adds: Highlight the corresponding field when focusing a document outline item

jcontonio:


jjidt (approved):
👍 looks good, this is a nice feature 💯

PR #755, opened 2020-08-17
A series of convenient updates

dheniges:
- Adds org id to API logging. Will make looking at active orgs much easier
- Updates link on Project API tokens to go to the API docs, instead of the docs homepage
- Removes API docs from the API, since the spec lives in portway-docs now


jjidt (approved):
👍

PR #757, opened 2020-08-17
update demoAccount invitation code to fix stripe error and update docs

dheniges:
Also cleans up the account cancellation language


jjidt (approved):
👍

PR #758, opened 2020-08-17
Removing notes language in the app

jcontonio:


web/src/server/views/partials/meta-tags.ejs
```diff
@@ -14,13 +14,13 @@
 <meta name="twitter:card" content="summary">
 <meta name="twitter:site" content="@portwayapp">
 <meta name="twitter:title" content="Portway">
-<meta name="twitter:description" content="A notes app for people who need to do more">
+<meta name="twitter:description" content="A writing app for people who need to do more">
```
dheniges: What do you think about "A collaborative writing app for people who need to do more"

jcontonio: Oh, yeah definitely

dheniges (approved):

PR #759, opened 2020-08-17
emit blur event with doc id as string

dheniges:
This was causing an error on document join room, since it wants a string, not a number. Also makes it consistent with the other events, which all seem to be passing strings

### DIFF
--- a/web/src/client/js/actions/userSync.js
+++ b/web/src/client/js/actions/userSync.js
@@ -7,6 +7,8 @@ export const emitJoinDocumentRoom = (documentId) => {
       return UserSync.socketError()
     }
     dispatch(UserSync.emitJoinDocumentRoom(documentId))
+    console.log('emitting joinRoom for ' + documentId)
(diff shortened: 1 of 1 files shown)


web/src/client/js/actions/userSync.js
```diff
@@ -7,6 +7,8 @@ export const emitJoinDocumentRoom = (documentId) => {
       return UserSync.socketError()
     }
     dispatch(UserSync.emitJoinDocumentRoom(documentId))
+    console.log('emitting joinRoom for ' + documentId)
```
jjidt: wanna leave these logs in there?

jjidt: ^

dheniges: oops no, will update here

dheniges:
@jjidt no more console

jjidt (approved):

PR #764, opened 2020-09-16
webhook delivery

dheniges:
Delivers webhooks


api/src/businesstime/webhook.js
```diff
@@ -0,0 +1,74 @@
+import ono from 'ono'
+
+import { getDb } from '../db/dbConnector'
+import resourceTypes from '../constants/resourceTypes'
+import resourcePublicFields from '../constants/resourcePublicFields'
+import { pick } from '../libs/utils'
+// import { getPaginationOptions, getSortOptions } from '../libs/queryFilters'
```
jjidt: leave ?

api/src/controllers/projectWebhooks.js
```diff
@@ -0,0 +1,117 @@
+import ono from 'ono'
+import Joi from '@hapi/joi'
+
+import { validateBody, validateParams, validateQuery } from '../libs/middleware/payloadValidation'
+import BusinessWebhook from '../businesstime/webhook'
+// import BusinessOrganization from '../businesstime/organization'
```
jjidt: need this?

api/src/db/models/webhookdelivery.js
```diff
@@ -0,0 +1,17 @@
+
+module.exports = (sequelize, DataTypes) => {
+  const WebhookDelivery = sequelize.define('WebhookDelivery', {
+    webhookId: DataTypes.INTEGER,
+    orgId: DataTypes.INTEGER,
+    resultCode: DataTypes.STRING
```
jjidt: is there a reason to have this as a string instead of an integer?

jjidt (changes requested):
couple commented out things and one question, tested it out with ngrok and a small express server and it delivered my webhooks :tada:

api/src/db/models/webhookdelivery.js
dheniges: in my head I wasn't sure if we'd have non-http response codes/strings, but maybe all webhooks will be http? None of the code currently cares whether it's a number or string. What do you think?

jjidt (approved):

PR #765, opened 2020-09-24
Webhooks Feature

dheniges:
TODO:
- [x] Webhook Icon
- [x] API docs
- [x] Cleanup job for old webhook deliveries
- [x] Enable/disable functionality in UI
- [x] Change placeholder in new hook field to point to a valid url `https://example.com/endpoint`


dheniges:
Things I'm noticing using this:
1. Cannot select url text
2. I think having a "Response Code" and "Date" label above the corresponding columns in the webhook delivery list would be useful
3. Fetching the first webhook delivery from the API is returning the oldest webhook delivery

jcontonio (approved):
🏴‍☠️

PR #768, opened 2020-10-06
Adding toggling / removing

jcontonio:


dheniges (approved):
Functionally works nicely 👍 

Style-wise, I dunno if you have any ideas for making it clearer the deliveries and the enable and delete buttons are nested under the webhook, they look a bit separate right now. 

Screen Shot 2020-10-08 at 3 49 20 PM

jcontonio:
Updated to make it more obvious. Also switched to a toggle

PR #778, opened 2020-10-29
kubernetes config updates for letsencrypt with cert-manager

dheniges:
Updated clusters to cert-manager v1, corresponding config updates in kube yaml


jcontonio (approved):
👍

PR #785, opened 2020-11-25
Sets a custom JSON Content-Type for assoc file

jcontonio:
Serves the associated domain file as JSON in Express


jjidt (approved):
let's try it out 👍

PR #788, opened 2020-11-26
Adds links to billing section in the emails that need it

jcontonio:
* Adds buttons to go to billing section in trial ending, trial ended, and payment failed
* Made some text smaller in some places

h/t Ashur


dheniges (approved):
I think this looks good, but we need to add an email test thing before we make any email changes in the future (after this PR). Too easy to screw up the templating and not realize it

PR #789, opened 2020-11-26
Fixes dark mode for the payment form

jcontonio:
* Does a React state change to get Stripes fields to accept new style values on dark mode change
* Also verified that 1Password can fill the form


dheniges (changes requested):
The country selector doesn't change, but everything else looks good

Screen Shot 2020-11-30 at 1 06 28 PM

jcontonio:
I see the country menu working in both Safari and Firefox. What browser you on? Shouldn't need a hard refresh either, my mistake.

dheniges:
it was Safari, but a hard refresh fixed it :idk:

dheniges (approved):

PR #790, opened 2020-11-30
Fixes payment expiration date display

jcontonio:
If a users card expired on a month <= 9, moment would have an invalid date due to the string we were making. https://github.com/moment/moment/issues/2554


dheniges (approved):
🥇 more dateful

PR #791, opened 2020-11-30
Fixes rendering bug in plan selector on Safari 14

jcontonio:
In Safari < 14, the plan selector green area (price) wouldn't go all the way across, so we had a hack in there.
Safari 14 fixed that bug, so it looked fucked up.


dheniges (approved):
It works in Safari 14. No idea what 13 looks like, it's above my pay grade to test

PR #799, opened 2020-12-09
All the deletions

dheniges:
- soft delete admin endpoint
- database constraints to remove records when foreign key references are deleted
- updated project deletion and document deletion to flow through the coordinators to handle asset deletion

Requires #779 first


api/src/businesstime/resourceusage.js
```diff
@@ -44,8 +44,22 @@ async function deleteAllForOrg(orgId, force = false) {
   })
 }
 
+async function deleteAllSoftDeletedBefore(timestamp) {
+  const db = getDb()
+
+  return db.model(MODEL_NAME).destroy({
+    where: {
+      deletedAt: {
+        [Op.lte]: timestamp
```
jjidt, line 55: Op needs imported in this file

api/src/controllers/admin.js
```diff
@@ -136,4 +148,26 @@ const deleteWebhookDelivery = async function(req, res, next) {
...
+    }
+    res.status(200).send({ data: { removedOrgsCount } })
+  } catch(e) {
+    next(e)
+  }
+}
+
+const deleteStaleResources = async function(req, res, next) {
+  try {
+    deleteSoftDeletedResources()
```
jjidt: need to await here, and then send a success status

jjidt (changes requested):

jjidt (approved):

PR #800, opened 2020-12-09
mark projects updated when fields and docs get C-U-D

jjidt:


jcontonio (approved):
Works 👍🏼 

We'll need to refresh the client updatedAt as well, as it takes a hard refresh to see the update. Do you want to do that in a separate PR?

PR #801, opened 2020-12-09
Fixes trial ended linkUrl bug

jcontonio:
We missed this one too, but that error in the email looks like it's from `sendTrialWillEnd`?


dheniges:
addressed this bug and fixed all the linkUrl emails in #804 https://github.com/BonkeyBong/portway/pull/804

PR #804, opened 2020-12-09
fix up emails with links to render correctly and add tests

dheniges:
the link variables needed to go to the ejs render functions, not the sendSingleRecipientEmail function. D'oh.


jcontonio (approved):
Looks good

PR #807, opened 2020-12-10
Web - temporary updatedAt on projects

jjidt:
- when docs and fields are created, updated, deleted, parent project gets updatedAt set in client data
- had to trigger an initial sort in the dashboard container, and remove a sort conditional in the projects reducer, which seems to be ok, but worth some thorough testing


jcontonio (approved):
Works well 👍🏼

PR #815, opened 2020-12-18
deploy feature branch

jjidt:


dheniges:
To deploy from a branch you need to create a release first and give it a tag. Creating the release does the image building, and then in this deployment PR it tells it what image tags to use.

When you create the release there's a branch selector, so you can choose the feature branch there, pick a tag, etc. Lemme know if you have questions.

jjidt:
updated to point to beta tag

dheniges (approved):
:realjj:

PR #816, opened 2020-12-19
set up per user plan config for dev

jjidt:


kubernetes/helm/portway/templates/api-deployment.yaml
```diff
@@ -94,6 +94,11 @@ spec:
             configMapKeyRef:
               name: api-config
               key: api.redis-url
+        - name: STRIPE_PER_USER_PLAN_ID
```
dheniges, line 97: should either make this optional or add a value in the prodValues yaml

jjidt:
made it optional, no prod plan created yet

dheniges:
checkout the docs/environment-variables.md file to make sure this gets documented and propagated around. Kube config side looks good

dheniges (approved):
:fakejj:

PR #821, opened 2020-12-22
Feature per user billing

jjidt:
- adds per user plan (id is via env STRIPE_PER_USER_PLAN_ID)
- new test setup file for API, which allows us to pass in test specific ENV
- removes plan selector from client
- new users get PER_USER plan
- if a SINGLE_USER adds seats, they get upgraded to PER_USER plan

To test:
on current release, make 3 accounts, and go through the steps to make:
1. a SINGLE_USER with billing info(TRIALING_PENDING_ACTIVE)
2. a MULTI_USER with billing info
3. keep one as a TRIALING account

change to this branch and test by adding billing info/adding seats for each user, and verify that their organization billing info and ORG get the new plan type

Then - create a user on the feature branch and go through each billing option


dheniges:
Screen Shot 2021-01-08 at 9 08 11 AM

The seats bar doesn't show the percentage of seats. Need to investigate

dheniges (approved):
It's ready to :shipit:

Some delete TODOs in the code which we can revisit someday when we purge ourselves of all previous subscription remnants.

PR #822, opened 2020-12-23
Remove team references and seat options based on plan and subscription status

jjidt:


dheniges:
I tested everything on this branch, and the issue I run into is existing single user plans (trial or paid) cannot upgrade to the new plan because they can't add seats.

jjidt:
updated, we were still only allowing MULTI_USER_PLANS to see the seat component

dheniges (approved):
Updates look good 👍

PR #829, opened 2021-01-05
send payment amount

jjidt:
updates payment success email with amount


dheniges (changes requested):
looks good, but can you update the coordinator test to pass a dollar amount and confirm it's in the body or something? It will render without an amount (well, as NaN), but trying to keep the email tests updated

jjidt:
updated with amount

dheniges (approved):

PR #831, opened 2021-01-08
WEB - mailing list endpoint (solidmuffins)

jjidt:


web/src/server/controllers/mailinglist.js
```diff
@@ -0,0 +1,24 @@
...
+
+const MailingListController = function (router) {
+  router.post('/', async (req, res) => {
+    const { email } = req.body
+
+    mailchimp.joinList(email)
+
+    // If we're getting hit from the external website, do a redirect
+    if (req.get('origin') === SUPPORT_FORM_SUBMIT_ORIGIN) {
+      const base = SUPPORT_FORM_SUBMIT_ORIGIN
```
jcontonio: We need to redirect to the page that submitted the form, instead of a specific page. Can you get the requesting domain / path and append the `?received=true` to that?

jjidt:
updated

jcontonio (commented):
I received an error when submitting the form, leaving me in a blank state on /solidmuffins. We should redirect even if there is an error. 

* If coming from ajax, respond with an error and error message
* If we're coming from a site, just redirect and silently fail I guess?

jcontonio (approved):

PR #833, opened 2021-01-11
Tweaks on new PER_USER components

jcontonio:
* Fixes meter tag when there is only 1 user and 1 seat
* Fixes some language on the users admin page when there is only 1 user / 1 seat
* Fixes linking between users page / billing page / project settings for TRIALING_PENDING_ACTIVE


jjidt (approved):
👍

PR #834, opened 2021-01-11
Fixes bundle for sign-up error page

jcontonio:


dheniges (approved):
I added a tweak to the signup js so it sets the disabled state based on the status of the checkbox. This fixes the issue where if you get to the page and the browser has the fields set to the last state you had them in (by using the back button for instance), the button will be correctly enabled or disabled

PR #835, opened 2021-01-16
fetch billing whenever seats change

jjidt:


dheniges (approved):
:+1: it does the thing

PR #840, opened 2021-01-27
Fixes responsive issues in the document view

jcontonio:
* Fixes scrolling in the document on mobile
* Fixes width of conflict resolution UI on mobile
* Adds resize handler to make sure the document toolbar is always correct


dheniges (approved):
the things work for me in FF and Safari

PR #844, opened 2021-01-29
Upgrades to new Stripe hooks; Adds payment route

jcontonio:
* Adds a new route under billing `/admin/billing/payment` that will open the credit card form
* On successful update, it will redirect to `/admin/billing` to close the credit card form
* Removed UI reducer for this stuff
* Upgrades to [React Stripe JS](https://github.com/stripe/react-stripe-js), so no more class components!


dheniges (dismissed):
No issues with functionality. The placeholder text coloring changed though so it's too hard to differentiate between text you entered and the placeholder:
Screen Shot 2021-02-02 at 10 44 31 AM

In the above screenshot, the state and zip code are placeholders, but the other fields are filled out.

Vs. the current styling:
Screen Shot 2021-02-02 at 10 44 36 AM

jcontonio:
@dheniges all set

jjidt (approved):
👍

PR #845, opened 2021-02-01
Feature doc refresh

jjidt:
- Organization sync
- Doc list refresh when other org users add docs
- upgrade socket.io to current major version


web/src/client/js/components/ProjectsList/ProjectsListContainer.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect } from 'react'
```
dheniges: 👀

dheniges (dismissed):
One extraneous import

jcontonio (approved):

PR #846, opened 2021-02-02
fix document header and field list flash when remote document created

jjidt:
- flattens document fields/header/panel components
- no more doc removal on sync creation event, which fixes doc header flash


web/src/client/js/components/Document/DocumentHeaderComponent.js
```diff
@@ -0,0 +1,162 @@
...
+  const panelButtonColor = documentMode === DOCUMENT_MODE.NORMAL ? 'transparent' : ''
+  const panelIconFill = documentMode === DOCUMENT_MODE.NORMAL ? 'var(--theme-icon-color)' : 'var(--color-blue)'
+
+  const changeHandlerAction = debounce(500, (e) => {
+    nameChangeHandler(e)
+  })
+
+  return (
+    <div>
+      <ValidationContainer resource="document" value="name" />
```
jcontonio: This can be removed right? It's in Project.js. Then that means this component should just be a <header> tag with no surrounding <div>

web/src/client/js/components/Document/DocumentComponent.js
```diff
@@ -145,7 +145,6 @@ const DocumentComponent = ({
           </div>
         </ProjectPermission>
       </header>
-      <DocumentFieldsContainer />
```
jcontonio, line 148: Remove both DocumentComponent and DocumentContainer

jcontonio (changes requested):
Looks good! Do you want to try merging `release` into this though? There are some mobile / responsive changes there that aren't here.

jcontonio (approved):
Merged in release, and fixed mobile navigating now that Project.js doesn't refresh all the time

PR #855, opened 2021-02-11
use the draft query for the draft endpoint, duh

jjidt:


jcontonio (approved):
Works for me, but @dheniges should check other functionality and make sure nothing breaks

PR #860, opened 2021-02-15
create magic field v. carefully

dheniges:
Steps to reproduce issue:
- Go to portway
- Navigate to a project. Do not select a document. Do not pass go. Do not collect $200
- Click in the blank document area
- Now click on a document
- The document will be disabled without this PR

Things this PR fixes:
- cannot invoke create field without a document ID and project ID
- Fixes click handler to add a field to an empty document by scoping it to document area only _and_ only fires when fields have been loaded and are empty


jjidt (commented):
Fixes the document disabled bug, will wait to see what we do with the chick handler and review again

dheniges:
going to update and add comments to that click handler, it should be triggered when there are no fields and a document and its fields are fully loaded. Hoping the code for that is as straightforward as the description 🤞

jjidt (approved):
👍 works great, tested multiple scenarios and seems to cover it (sidebar open, sidebar closed, navigating from other pages etc.)

PR #863, opened 2021-02-16
payment email amount bugfix

jjidt:
handles payment that aren't rounded dollar amounts


dheniges (approved):
Looked at the webhook payload from stripe and added a test for the email generation. Looks good 👍

jjidt:
oh yeah, tests, thanks dirk

PR #865, opened 2021-02-22
option to set markUpdated on project delete

dheniges:
Since the markUpdated() is async and not awaited, it can trigger after the project or document has been deleted. When deleting a project it makes a bunch of unnecessary updates (because the resources are being deleted!). Added option to override the "markUpdated" functionality to all the functions.


jjidt (approved):
👍

PR #867, opened 2021-02-23
delete stripe customer when deleting unverified orgs

dheniges:
Lots of 404s on stripe webhooks and they keep retrying because we deleted the unverified users but not the stripe customer

Oops branched off #865 so that should get reviewed first


jjidt (changes requested):
When a customer manually cancels their account, they no longer have a subscription and we get a 500 error in the billing coordinator's formatBilling function

I think we need an existence check on the billingSubscription on line 44 there.

`const portwayPlanId = billingSubscription ? STRIPE_PLAN_ID_TO_PORTWAY_PLAN_MAP[billingSubscription.plan.id] : null`

jjidt (approved):

PR #868, opened 2021-02-23
Remove worker markdown processing

dheniges:
- Removes the worker thread task runner code
- Removes the worker thread markdown processing
- Adds logging for markdown processing where value does not match structuredValue


jjidt (approved):
👍

PR #869, opened 2021-02-23
image conversion job

jjidt:
- Adds an image conversion worker 
- Currently just logs the results to the console so you can get the processed-image urls
- Attempted to do this with streaming, but sharp loads the data into a buffer to extract metadata, so that wasn't giving us any advantage with memory (also it was buggy)

Next Steps:
- create field structure to return formats
- update field after job runs (the job will probably need an admin route to update fields), I'm currently passing the document and field id through the job queue to make this happen
- strip metadata on the original image, can use sharp library on the api to do this
- attach our metadata to the field item


workers/src/coordinators/imageProcessing.js
```diff
@@ -0,0 +1,56 @@
...
+
+  // webPHalf
+  const webPHalf = await image
+    .resize(Math.round(metadata.width / 2))
+    .toFormat('webp')
+    .toBuffer()
+
+  // pngFull
+  const pngFull = await image
+    .toFormat('png')
```
dheniges: I think we are going with `avif` not `png` as the 2nd format

jjidt:
updated

workers/src/coordinators/imageProcessing.js
```diff
@@ -0,0 +1,56 @@
...
+    .toBuffer()
+
+  // webPHalf
+  const webPHalf = await image
+    .resize(Math.round(metadata.width / 2))
+    .toFormat('webp')
+    .toBuffer()
+
+  // pngFull
+  const pngFull = await image
```
jcontonio: Change var names to avif

jjidt:
updated

jcontonio (approved):

PR #874, opened 2021-02-24
Optional subscription handling

jjidt:
handle cases where user cancels account and no longer has a subscription


dheniges (approved):
after careful code review, :jjidt:

dheniges:
:idt:

PR #885, opened 2021-03-09
image field metadata

jjidt:
- adds original width and height to field meta object when uploading image
- adds alignment and alt field props, settable via api(we'll want to make 'center' default in our UI, not in the API)
- uses special getters to not return anything if alt or alignment are null


dheniges (approved):
:realjj:

PR #888, opened 2021-03-10
Adds new image settings view

jcontonio:
**Contains all of JJ's code, review and merge that first**

* Shows webp versions of images if they exist
* Updates alt text and alignment on image fields
* Allows uploading of images even when not in settings mode
* Fixes iPad portrait mode document view


jjidt (changes requested):
Code looks good, I'm noticing some odd behavior when changing some of the new settings on an existing image, and then uploading a new image or two without clicking the checkmark

It occasionally jumps back to the first image, or doesn't change

jjidt:
@jcontonio take a look at the issue above whenever you have time

jcontonio:
@jjidt want to try this again? Before, uploading the image would set the preview, but the srcSet and webpSource were still there from the previous field values.

jjidt (approved):
nice!

PR #889, opened 2021-03-11
always show team regardless of org read/write settings

jjidt:
@jcontonio check this out and see if it's working they way it should


dheniges (approved):
seems fine for all the permission levels 👍

PR #890, opened 2021-03-16
add renderedValue to fields

jjidt:
- just dumb ejs templates for now, can hand this off to Jay to expand on those
- there's a todo in there for passing along the image alt, width, and height when those are available


jcontonio:
Here's what I'm thinking for each of the templates:

1. Add `type`, `name`, and `value` to all the fields, and use this as the generic template for everything. This way someone can style Portway divs specifically, and then style field types individually

```ejs
<div data-source="portway" data-type="<%= type %>" data-name="<%= name %>">
	{{ value }}
</div>
```

2. For the File template, add `download` attribute:

```ejs
<div data-source="portway" data-type="<%= type %>" data-name="<%= name %>">
	<a href="<%= value %>" download><%= name %></a>
</div>
```

3. For the Date template, use the `time` [tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time). Let the debate begin on the format here:

```ejs
<div data-source="portway" data-type="<%= type %>" data-name="<%= name %>">
	<time datetime="<%= value %>"><%= formattedDate? %></time>
</div>
```

4. For the Image template, we'll want to add the srcset, and add an alignment attribute:

```ejs
<div data-source="portway" data-type="<%= type %>" data-name="<%= name %>" data-align="<%= alignment %>">
	<img
		src="<%= webpSource.half %>"
		srcSet="<%= webpSource.half %>, <%= webpSource.full %> 2x"
		alt="<%= alt %>"
		width="<%= width %>"
		height="<%= height %>"
		align="<%= alignment %>"
	/>
</div>
```

jjidt:
need to update renderedValue when image conversion job is done

jjidt:
- updated with a publishing fix (all fields get copied over for a published field, rather than cherry picking like we were before), so publishing reliability should be verified for all field types
- plugged in new templates, not sure if we want all the escaping that is happening, also image fields will briefly have empty srcSet values until the worker is done processing, not sure if that will have side effects

jcontonio:
@jjidt whenever you're done with duplicate let's remove whatever escaping is going on here, then we're good

api/src/coordinators/markdown.js
```diff
@@ -1,9 +1,13 @@
 import MarkdownIt from 'markdown-it'
 import processTokens from '../libs/markdownProcessor'
 
-const md = new MarkdownIt()
+const md = new MarkdownIt({ breaks: true })
```
dheniges, line 4: This seems like it would impact the structured value output if it's sticking `br`s on all newlines, but the tests still pass so maybe it's not?

api/src/coordinators/documentVersion.js
```diff
@@ -91,7 +89,10 @@ const processNewFieldVersionByType = function(field, newField) {
       newField.structuredValue = field.structuredValue
     }
   }
-  return newField
+  const fieldForUpdate = { ...field, ...newField }
+  delete fieldForUpdate.documentId
```
dheniges, line 93: `documentId` doesn't change here does it?

api/src/coordinators/documentVersion.js
```diff
@@ -91,7 +89,10 @@ const processNewFieldVersionByType = function(field, newField) {
...
+  delete fieldForUpdate.id
+  return fieldForUpdate
```
dheniges, line 95: This is going from a whitelist to a blacklist of fields, but still has to deal with the individual case of the structuredValue. Guess it's a tradeoff between forgetting to update this function with new fields but still having them in the API, vs. exposing internals we didn't mean to. :shrug:

api/src/businesstime/field.js
```diff
@@ -57,6 +58,12 @@ async function createForDocument(documentId, body) {
 
   await createdField.setFieldValue(fieldValue.id)
 
+  // set the rendered value on the field
+  if (fieldValue.value != null) {
+    const renderedValue = await getRenderedValueByType(createdField, fieldValue.value)
+    await createdField.update({ renderedValue })
```
dheniges: Doesn't this make more sense in the field coordinator? It would save a database hit (createdField created, then subsequently updated here for the renderedValue)

api/src/businesstime/field.js
```diff
@@ -159,6 +166,12 @@ async function updateByIdForDocument(id, documentId, orgId, body) {
 
   await fieldValue.update({ value: body.value, structuredValue: body.structuredValue })
 
+  // set the rendered value on the field
+  if (fieldValue.get('value') != null ) {
+    const renderedValue = await getRenderedValueByType(updatedField, fieldValue.value)
+    await updatedField.update({ renderedValue })
```
dheniges: Same comment, I could be missing something but this seems like a coordinator action

api/src/libs/fieldRenderedValue/index.js
```diff
@@ -0,0 +1,64 @@
...
+      async: true // returns an async render func
+    }
+  )
+})
+
+/**
+ * Takes a field body and returns the rendered value for each type of field
+ * @param {FieldBody} field -
+ */
+export function getRenderedValueByType(field, value) {
```
dheniges, line 34: Seems like throwing a few tests here for different field types is worth the time. I also think a warning log if the input field type isn't found might save us a future headache if we forget to update this with a new field type in the future.

dheniges (changes requested):
Some changes to make, but it worked for me 👍 

Let me know if I'm missing something on the renderedValue being computed in the Business layer.

api/src/coordinators/markdown.js
jjidt: I think I checked and the structured value was already doing this by default, we can double check though

dheniges: peeked at the tests and they definitely cover it, so we're good

api/src/coordinators/documentVersion.js
jjidt: yeah, if you have a suggestion for a way you'd prefer I'm open to it, if it's a whitelist, then adding a field would require updating here, updating the incoming duplication code, and updating the public fields array, but maybe if they were all constants in the same file that would be easier to remember

dheniges: I like the idea of defining it in constants, do you want to add that in asana?

api/src/coordinators/documentVersion.js
jjidt: it doesn't, but we'll get an error if we try to pass that on the update

api/src/businesstime/field.js
jjidt: I had that originally, but moved it here because we needed additional fields that weren't coming in on the body

https://github.com/BonkeyBong/portway/pull/890/commits/5ab542842cf752191a1baf27ec1e837d599cde05

jjidt:
updated, rendering is now happening in the coordinator, and lots of fancy new tests

dheniges (approved):

PR #891, opened 2021-03-22
document deletion webhook

jjidt:


jcontonio (approved):
Deleted a document, it webhooked 👍🏼

PR #896, opened 2021-03-31
document duplicate

jjidt:
- document duplicationz with tests
- after this is merged, we can convert the intro project coordinator to use the same field duplication

dupe endpoint is:

`/projects/:projectId/documents/:docId/duplicate`

 Next steps: update the documentation with new endpoint

### DIFF
--- a/api/src/businesstime/field.js
+++ b/api/src/businesstime/field.js
@@ -45,6 +45,7 @@ async function createForDocument(documentId, body) {
 
   // Debugging log for mismatched text value to structuredValue
   // Added 2/24/2021 -DH
+
   if (body.type === FIELD_TYPES.TEXT && body.value && !Boolean(body.structuredValue)) {
     logger(`DEBUG: found TEXT field with length ${body.value.length} with empty structuredValue in BusinessField.createForDocument`)
   }
--- a/api/src/businesstime/field.js
+++ b/api/src/businesstime/field.js
@@ -45,7 +45,6 @@ async function createForDocument(documentId, body) {
 
   // Debugging log for mismatched text value to structuredValue
   // Added 2/24/2021 -DH
-
   if (body.type === FIELD_TYPES.TEXT && body.value && !Boolean(body.structuredValue)) {
     logger(`DEBUG: found TEXT field with length ${body.value.length} with empty structuredValue in BusinessField.createForDocument`)
   }
--- a/api/src/coordinators/document.js
+++ b/api/src/coordinators/document.js
@@ -56,10 +56,24 @@ const deleteAllForProject = async (projectId, orgId) => {
   }))
 }
 
+const duplicateDocument = async (documentId, projectId, orgId) => {
+  // duplicate the document
+  const dupDoc = await BusinessDocument.duplicateById(documentId, orgId)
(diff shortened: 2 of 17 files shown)


api/src/businesstime/field.js
```diff
@@ -45,6 +45,7 @@ async function createForDocument(documentId, body) {
 
   // Debugging log for mismatched text value to structuredValue
   // Added 2/24/2021 -DH
+
   if (body.type === FIELD_TYPES.TEXT && body.value && !Boolean(body.structuredValue)) {
```
dheniges:
```suggestion
  // Added 2/24/2021 -DH
  if (body.type === FIELD_TYPES.TEXT && body.value && !Boolean(body.structuredValue)) {
```

api/src/coordinators/document.js
```diff
@@ -56,10 +56,34 @@ const deleteAllForProject = async (projectId, orgId) => {
...
+    orgId
+  }
+  // create the duplicate document
+  const dupeDoc = await BusinessDocument.createForProject(projectId, body)
+  // loop through fields and create the duplicates
+  await Promise.all(fields.map((field) => {
+    return FieldCoordinator.duplicateField(field.id, document.id, dupeDoc.id, orgId)
+  }))
+  // we could technically assemble this payload from the retuned values above, but
+  // just in case, go through the normal pathway for fetching
```
dheniges, line 78: :+1:

api/src/coordinators/field.js
```diff
@@ -129,17 +129,39 @@ const getFieldBodyByType = async function(body, documentId, orgId, file) {
...
+  const field = await BusinessField.findByIdForDocument(id, originalParentDocId, orgId)
+
+  const body = FIELD_PROPS_TO_COPY.reduce((body, fieldProp) => {
+    body[fieldProp] = field[fieldProp]
+    return body
+  }, {})
+  body.orgId = orgId
+
+  if (body.type === FIELD_TYPES.IMAGE) {
+    await fieldCoordinator.addImageFieldFromUrlToDocument(newParentDocId, body, body.value)
```
dheniges: I think this needs to return, otherwise it creates two image fields

api/src/coordinators/document.js
```diff
@@ -56,10 +56,34 @@ const deleteAllForProject = async (projectId, orgId) => {
...
+  const body = {
+    name: `${document.name}_copy`,
+    slug: `${document.slug}_copy`,
+    projectId: document.projectId,
+    orgId
+  }
+  // create the duplicate document
+  const dupeDoc = await BusinessDocument.createForProject(projectId, body)
+  // loop through fields and create the duplicates
+  await Promise.all(fields.map((field) => {
```
dheniges: this runs async, so order isn't preserved. I didn't trace it back, but I'm sure somewhere the `normalizeFieldOrder...` function or whatever it's called is getting called. Maybe in the image copy on field coordinator? Not sure

dheniges (changes requested):
It's close, a couple issues:
- Double image trouble
- Order preservation

jjidt:
nice catches there, updated

dheniges (approved):

PR #909, opened 2021-04-19
shared field props for copying

jjidt:
- intro and duplication now have a shared constant containing the props to copy
- fixes structuredValue copying for the intro

### DIFF
--- a/api/src/constants/fieldTypes.js
+++ b/api/src/constants/fieldTypes.js
@@ -27,4 +27,6 @@ export const IMAGE_ALIGNMENT_OPTIONS = {
   RIGHT: 'right'
 }
 
+export const FIELD_PROPS_TO_COPY = ['type', 'value', 'order', 'name', 'structuredValue']
+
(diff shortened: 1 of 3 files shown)


api/src/constants/fieldTypes.js
```diff
@@ -27,4 +27,6 @@ export const IMAGE_ALIGNMENT_OPTIONS = {
   RIGHT: 'right'
 }
 
+export const FIELD_PROPS_TO_COPY = ['type', 'value', 'order', 'name', 'structuredValue']
```
jcontonio: Shouldn't we copy everything? The images will have alt, alignment, formats, etc?

jcontonio (approved):
Copied over some structuredValues, but not all. Assuming that's the other structuredValue bug

Screen Shot 2021-04-20 at 12 11 08 PM

jjidt:
good catch @jcontonio, added the additional field props

PR #910, opened 2021-04-20
Adds status for duplicating documents

jcontonio:
* Adds a new StatusSpinner component that can show a string and a spinner left or right
* Adds a Status state to UI reducer
* Shows a status in the ProjectToolbar and disables duplication if we are duplicating a document
* Can use this status for other things in the future


jjidt (approved):
works great 👍

PR #917, opened 2021-04-23
Adds timed countdown of maxLength in the string field

jcontonio:
Type in a string field and see the counter show up for 5 seconds


jjidt (approved):
nice 👍

PR #918, opened 2021-04-26
Fixes document slug client validation

jcontonio:
Typing an invalid slug value should show an error message in all the browsers
(don't merge this until DocumentPanel is merged)


jjidt (approved):
looks good 👍

PR #919, opened 2021-04-26
doc name fix

jjidt:
- tracks document name in doc header using internal state
- only updates from the api if not the currently focused field


jcontonio:
Works from the panel side... is there any way to get it to work vice versa? I edit the header of the doc, but the name input doesn't change in the panel

jjidt:
updated

jcontonio (approved):
Hooray 👍🏼

PR #920, opened 2021-04-27
fix field update validation/fix meta bug

jjidt:
- there was a bug with image fields where meta didn't exist, fixed here
- also field update validation should have been checking the original update body rather than the one we create in the coordinator, not sure how that got reverted, but also fixed here


dheniges (approved):
:+1:

dheniges (changes requested):
sorry revoking approval, missed that tests weren't passing

dheniges (approved):
huzzah

PR #921, opened 2021-04-27
Changes document footer

jcontonio:
* The un-publish and delete buttons move to the document info panel
* The document footer no longer disables when the panel is open


jjidt (approved):
nice 👍

PR #925, opened 2021-04-28
Fixes file extension bug in field settings

jcontonio:


jjidt (changes requested):
we need to do the same thing for file size in the settings component, getting the same error for file fields with nothing uploaded

probably also want to disable download when there's nothing there yet

jjidt (approved):

PR #927, opened 2021-04-28
always truncate image duplication filenames to 14 characters

jjidt:
fixes copying a copy of a copy of a copy of a copy bug

### DIFF
--- a/api/src/coordinators/field.js
+++ b/api/src/coordinators/field.js
@@ -51,13 +51,14 @@ const addImageFieldFromUrlToDocument = async function(documentId, body, url) {
 
   const fileStats = await callFuncWithArgs(stat, filePath)
 
-  const urlParts = url.split('/')
-  const name = urlParts[urlParts.length - 1]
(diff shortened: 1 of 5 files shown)


api/src/coordinators/field.js
```diff
@@ -51,13 +51,14 @@ const addImageFieldFromUrlToDocument = async function(documentId, body, url) {
 
   const fileStats = await callFuncWithArgs(stat, filePath)
 
-  const urlParts = url.split('/')
-  const name = urlParts[urlParts.length - 1]
+  const urlPath = path.parse(url)
+  // truncate the file name
+  const truncatedName = `${urlPath.name.slice(0, 14)}${urlPath.ext}`
```
jcontonio: Should we do this for file fields as well?

jcontonio (commented):
Copies and publishes fine. Just the one question on File field names

jcontonio (approved):

PR #937, opened 2021-05-07
ensure body.value before running additional processes on field value

jjidt:
we were getting 500s when trying to duplicate documents with an empty image field, this add an additional check for a current body.value url before doing the duplication

### DIFF
--- a/api/src/coordinators/field.js
+++ b/api/src/coordinators/field.js
@@ -183,10 +183,10 @@ const duplicateField = async function(id, originalParentDocId, newParentDocId, o
     return body
   }, {})
   body.orgId = orgId
-
-  if (body.type === FIELD_TYPES.IMAGE || body.type === FIELD_TYPES.FILE) {
+  console.log(body)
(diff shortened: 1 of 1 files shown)


api/src/coordinators/field.js
```diff
@@ -183,10 +183,10 @@ const duplicateField = async function(id, originalParentDocId, newParentDocId, o
     return body
   }, {})
   body.orgId = orgId
-
-  if (body.type === FIELD_TYPES.IMAGE || body.type === FIELD_TYPES.FILE) {
+  console.log(body)
```
jcontonio: 👀

jcontonio (commented):
Remove that and 👍🏼

api/src/coordinators/field.js
jjidt: oops

jjidt: updated

jcontonio (approved):

PR #947, opened 2021-05-13
Changes responsive image to picture tag with fallback for webp

jcontonio:
* Uses `<picture>` tag everywhere for responsive images, falling back to original format if webp is not supported

@jjidt needs:

1. Add mime-type to each image format:
    ```javascript
    formats: {
      webp: {
        original: ...,
        half: ...,
        mimeType: ....
      }
    }
    ```
2. Add mime-type to the `type` attribute in the three spots marked `<JJ_HERE>`
3. Add a conditional in each case to display SVGs as normal img tags, not in a picture tag with any formats


jjidt:
updated, check it out

jjidt (approved):

